PR #106, opened 2015-06-15
copy path together with an example

lobingera:
As the title tells you, implementation to copy_path and copy_path_flat calls and a local convert function from cairo data to a reasonable julia data format (Array of entries of command type and points). This is done via some understanding of the underlying c struct and pointer arithmetic unsafe_load.


src/Cairo.jl
```diff
@@ -511,6 +514,86 @@ end
...
+    path_data = CairoPathEntry[]
+    data_index = 1
+    while data_index <= ((c.num_data)*2) 
+
+        # read header
+        element_length = reinterpret(Uint64,unsafe_load(c.data,data_index)) >> 32
+        element_type = reinterpret(Uint64,unsafe_load(c.data,data_index)) & 0xffffffff
+
+        # read points
+        points = Array(Float64,(element_length-1)*2)
```
timholy: Can you just use `pointer_to_array` here?

samples/sample_copy_path.jl
```diff
@@ -0,0 +1,73 @@
+
+using Cairo
+
+function common_header()
+	## header to provide surface and context
```
timholy: Indentation is unconventional in this file

src/Cairo.jl
```diff
@@ -511,6 +514,86 @@ end
...
+    finalizer(path, destroy)
+    path
+end    
+
+function convert_cairo_path_data(p::CairoPath)
+    c = unsafe_load(p.ptr)
+
+    # As you already can guess from the unsafe_load, we do here some pointer
+    # arithmetic and lowlevel data type stuff. In a better world, julia would
+    # provide a straight-forward counterpart to c structs...
```
timholy: Can't you use an `immutable` for such things?

timholy:
I don't have independent knowledge of these functions, but overall this looks very thorough.

samples/sample_copy_path.jl
lobingera: indentation is done, how sublime2 does it. What is the convention on indentation?

src/Cairo.jl
lobingera: Where?
I somehow had to solve the problem, that in the c world the entries are a union of two doubles OR two int32 (with padding). And copying is needed anyway.

src/Cairo.jl
lobingera: Specify "things".

timholy:
This is great, thanks!

Issue #107, opened 2015-06-23
Cairo scripting surfaces


lobingera:
libcairo brings some performance tools (opt/src/cairo-1.12.18/perf) which work on recorded traces of cairo calls. As julia is dynamically (at runtime) opening the libcairo, the cairo-trace tool, which replaces the libcairo at loading time with a library that traces and writes out the calls to a file, cannot be used.

So, Cairo.jl needs the CairoScriptSurface family of calls to enable writing calls to a trace.

lobingera:
so you could do this:

```
lobi@orange4:~/juliarepo$ ../julia/julia 
               _
   _       _ _(_)_     |  A fresh approach to technical computing
  (_)     | (_) (_)    |  Documentation: http://docs.julialang.org
   _ _   _| |_  __ _   |  Type "help()" for help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 0.3.10-pre+4 (2015-06-04 13:19 UTC)
 _/ |\__'_|_|_|\__'_|  |  Commit c8804e8* (19 days old release-0.3)
|__/                   |  x86_64-linux-gnu


shell> cat test_speed1.jl
include("shape_functions.jl")

s_size = 1024

s = Cairo.CairoARGBSurface(s_size,s_size);
c = Cairo.CairoContext(s);

@time rdots2(c,s_size,s_size,3.0,10000)

s = Cairo.CairoScriptSurface("rdots2.sc",s_size,s_size);
c = Cairo.CairoContext(s);
rdots2(c,s_size,s_size,3.0,10000)
Cairo.flush(s)

julia> include("test_speed1.jl")
elapsed time: 0.218702488 seconds (334288 bytes allocated)


shell> cairo-perf-trace rdots2.sc
[ # ]  backend                         test   min(s) median(s) stddev. count
[  0]      xcb                       rdots2    0.169    0.170  14.14%   15/15
[  0]     xlib                       rdots2    0.176    0.183  24.10%   15/15
[ # ]    image: pixman 0.30.2
[  0]    image                       rdots2    0.201    0.201   0.03%    5/6
[ # ]  image16: pixman 0.30.2
[  0]  image16                       rdots2    0.201    0.201   0.05%    6/6

```

lobingera:
PR close, i need to (re-)create master on my githup repository.

PR #108, opened 2015-06-23
Cairo script

lobingera:
#107 

I'm already using this for some benchmarking. The commit history looks strange - as i was using master for some other 1.12 features. The last (239a146) should actually be enough.


lobingera:
This is still valid, but i'll reognize this together with other features.

PR #114, opened 2015-08-27
Add pattern_get_surface function.

dcjones:
Adds bindings for the [`cairo_pattern_get_surface`](http://cairographics.org/manual/cairo-cairo-pattern-t.html#cairo-pattern-get-surface) function, which I found myself needing.


timholy:
Add a test?

lobingera:
@dcjones: Are you working on bitmap in Compose.jl? I ask, because i do also and could solve the cairo part (getting the data into the correct place and scale the bitmap), but failed on getting the coordinates correctly from prim type.

dcjones:
Test added.

@lobingera This is actually for something we discussed in https://github.com/dcjones/Gadfly.jl/issues/555, which was rendering repeated shapes (like in scatter plots) by drawing to a bitmap and and copying it in place repeatedly. This optimization get applied automatically and makes scatter plots in Gadfly about 8x faster on Cairo. You can try it out on the `measures` branch of Compose/Gadfly.

lobingera:
I see. I opened  dcjones/Compose.jl#140.

dcjones:
Ok, finally got this right.

timholy:
Thanks!

PR #117, opened 2015-08-30
Implement width and height for CairoContext

timholy:
Previously we had (with the new tests)
```jl
Cairo.CairoContext must implement width
 in width at no file
 in anonymous at test.jl:90
 in do_test at test.jl:50
 in include at ./boot.jl:259
 in include_from_node1 at ./loading.jl:267
 in process_options at ./client.jl:308
 in _start at ./client.jl:411
 in width at no file
 in anonymous at test.jl:90
 in do_test at test.jl:50
 in include at ./boot.jl:259
 in include_from_node1 at ./loading.jl:267
 in process_options at ./client.jl:308
 in _start at ./client.jl:411
while loading /home/tim/.julia/v0.3/Cairo/test/runtests.jl, in expression starting on line 8
```
so Cairo was not living up to the promises made by Graphics.

### DIFF
--- a/src/Cairo.jl
+++ b/src/Cairo.jl
@@ -353,6 +353,9 @@ function destroy(ctx::CairoContext)
     nothing
 end
 
+ width(ctx::CairoContext) =  width(ctx.surface)
+height(ctx::CairoContext) = height(ctx.surface)
+
 function copy(ctx::CairoContext)
     surf = surface_create_similar(ctx.surface)
     c = creategc(surf)
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,9 +1,16 @@
+using Cairo
+using Base.Test: @test, @test_throws
+
+surf = CairoImageSurface(100, 200, Cairo.FORMAT_ARGB32)
+@test width(surf) == 100
+@test height(surf) == 200
+ctx = CairoContext(surf)
+@test width(ctx) == 100
+@test height(ctx) == 200
+
 include("shape_functions.jl")
 include("test_stream.jl")
 include("tex.jl")
-
-using Base.Test: @test, @test_throws
-
 
 function test_pattern_get_surface()
     # test getting a surface from a surface pattern


lobingera:
Let's say this is breaking a little bit the expectations of libcairo users. Context has no size, the underlying surface has. And in special cases like recording surface, even this is missing.

timholy:
In this package there is no way to [create a `CairoContext`](https://github.com/JuliaLang/Cairo.jl/blob/0f607a417960fa6968c3ba4e95b1ea32d57c956a/src/Cairo.jl#L333-L341) without having a surface defined.. There's also [this in Graphics](https://github.com/JuliaLang/Graphics.jl/blob/208d433662b07065e929ca94d4a896bb96021827/src/Graphics.jl#L204-L205).

That said, I suspect you have more experience with Cairo than me. Should you decide that some cleanups are necessary, by all means go for it.

Relatedly, do you think we should move this package to JuliaGraphics?

lobingera:
@timholy.

For your question to move Cairo.jl to JuliaGraphics; well, it sounds like a match, but then GUIs like Tk and Gtk should show up there also.

For the comment about my experience with libcairo in other places: i was for some time now a user of pygtk/pycairo and i can write programs in C using libcairo (and i also spend some time in the cairo code to understand e.g. performance issues). At the same time i still do not have a clear understanding how packages like Gtk and Cairo should provide interfaces. Low-level and just as language adaption or rather high-level providing matching julia constructs. In pycairo e.g. cairo context is an object and all path construction functions are methods on the object. Which is quite good readable and if you find something in the libcairo C-API, so find the pycairo construct in seconds. 
A julia adaptation with this multi-dispatch thinking can act more freely. You can call the function to setup a surface with the C-API parameters or (i think we have that example) call the constructor with a filename and setup the surface and read already a picture into is.

On the general structure of Cairo.jl, i'd like to see more 1.12 or 1.14 readyness, some restructuring to move the text setting - which actually is pango/pangocairo - to a dedicated part. And there should be more doc strings...
To have both data structures, a GraphicsContext that is more abstract, providing more data than just the plain CairoContext could be an option - but this should also be driven by having a second example of a drawing/rendering toolkit.

And for the starter, the context without surface: Gtk does provide a ready-to-use cairo context for expose/draw calls together with a clip path. You see this actually showing up in the Gtk.jl/src/cairo.jl for the canvas draw calls.
I'm using locally a fork of Gtk and Cairo.jl that enables using this without the intermediate step of drawing to the back surface. For my (pending and slow moving) interactive Gtk example i use this.

PR #119, opened 2015-09-13
CairoContext from pointer only,

lobingera:
In working with Gtk (but not limited to) it's possible to get in calls or callbacks a cairo context as single pointer. This extends to the CairoContext by getting the surface information and adding a pango layout.


timholy:
Looks fine. I've not had this come up, however, is it possible to give a quick sketch of when you need this?

lobingera:
'when' as in: Giving a certain date? 'when' as in: Specify the need?

for the first: i'm using a local branch with this since ~feb.
for the second: i've mentioned this here: https://github.com/JuliaLang/Gtk.jl/pull/170#issuecomment-139213820; using a own draw method in Gtk returns at call the widget, user specific data and the cairo context, organized by Gtk.

PR #123, opened 2015-10-11
To reduce deprecation warnings: some whitespace and Libc for strftime

lobingera:
The title more or less says it, you got quite many deprecation warnings running the examples on a 0.4, as the programs copied literally from c-code and had some whitespace in the function call.


lobingera:
Is there a problem with this?

timholy:
Nope, just that no one noticed. Thanks!

PR #127, opened 2015-12-03
better error message when bindeps fails

tkelman:
suggest Pkg.build if deps.jl was not created successfully by BinDeps

### DIFF
--- a/src/Cairo.jl
+++ b/src/Cairo.jl
@@ -4,7 +4,9 @@ module Cairo
 
 using Compat
 
-include("../deps/deps.jl")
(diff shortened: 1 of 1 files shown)


src/Cairo.jl
```diff
@@ -4,7 +4,9 @@ module Cairo
 
 using Compat
 
-include("../deps/deps.jl")
+depsjl = joinpath(dirname(@__FILE__), "..", "deps", "deps.jl")
```
timholy: You know better than I, but does `".."` work on Windows too?

tkelman: yeah

tkelman: at least in include. not sure about dirname, isfile, etc

PR #129, opened 2015-12-21
adding zypper support

colbec:


nalimilan:
The failure is due to the fact that no released BinDeps version exports `Zypper` yet. But the fix looks good to me it if works for you -- except that you'll have to require BinDeps 0.3.21 once it's released.

colbec:
Got it, thankyou. I will hold fire on this, nothing urgent.

REQUIRE
```diff
@@ -1,7 +1,7 @@
 julia 0.3
 Compat 0.3.5
 Colors
-BinDeps
+BinDeps 0.3.20
```
tkelman: This can be rebased now that 0.3.21 was tagged.

timholy:
Thanks!

PR #131, opened 2016-01-26
Add paint_with_alpha()

cormullion:
I'm trying to add paint_with_alpha to Cairo.jl. This addition appears to work. Not sure how to add tests, but this works:

    ## header to provide surface and context
    using Cairo
    c = CairoRGBSurface(256,256)
    cr = CairoContext(c)

    save(cr)
    set_source_rgb(cr, 0.8, 0.8, 0.8)    # light gray
    rectangle(cr, 0.0, 0.0, 256.0, 256.0) # background
    fill(cr)
    restore(cr)

    save(cr)
    image = read_from_png(Pkg.dir("Cairo", "samples/data/", "mulberry.png"))
    w = image.width
    h = image.height

    translate(cr, 128.0, 128.0)
    rotate(cr, 45* pi/180)
    scale(cr, 256.0/w, 256.0/h)
    translate(cr, -0.5*w, -0.5*h)

    set_source_surface(cr, image, 0, 0)
    paint(cr)
    restore(cr)

    save(cr)
    scale(cr, 256.0/w, 256.0/h)
    set_source_surface(cr, image, 0, 0)
    paint_with_alpha(cr)                         # defaults to 0.5
    restore(cr)

    ## mark picture with current date
    move_to(cr, 0.0, 12.0)
    set_source_rgb(cr, 0, 0, 0)
    show_text(cr, Libc.strftime(time()))
    write_to_png(c, "sample_image.png")

sample_image

### DIFF
--- a/src/Cairo.jl
+++ b/src/Cairo.jl
@@ -438,11 +438,9 @@ function stroke_preserve(ctx::CairoContext)
     restore(ctx)
 end
 
-function paint_with_alpha(ctx::CairoContext, a = 0.5)
+function paint_with_alpha(ctx::CairoContext, a)
     ccall((:cairo_paint_with_alpha, _jl_libcairo),
-        Void,
-        (Ptr{Void}, Float64),
-        ctx.ptr, a)
+          Void, (Ptr{Void}, Float64), ctx.ptr, a)
 end
 
 function get_operator(ctx::CairoContext)
--- a/src/Cairo.jl
+++ b/src/Cairo.jl
@@ -55,7 +55,7 @@ export
     clip, clip_preserve, reset_clip,
 
     # fill, stroke, path, and shape commands
-    fill, fill_preserve, new_path, new_sub_path, close_path, paint, stroke,
+    fill, fill_preserve, new_path, new_sub_path, close_path, paint, paint_with_alpha, stroke,
     stroke_preserve, stroke_transformed, stroke_transformed_preserve,
     move_to, line_to, rel_line_to, rel_move_to,
     rectangle, circle, arc, arc_negative,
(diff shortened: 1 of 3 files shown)


src/Cairo.jl
```diff
@@ -438,6 +438,13 @@ function stroke_preserve(ctx::CairoContext)
     restore(ctx)
 end
 
+function paint_with_alpha(ctx::CairoContext, a = 0.5)
+    ccall((:cairo_paint_with_alpha, _jl_libcairo),
+        Void,
```
nalimilan: Please align this line and the following ones to the second parenthesis.

src/Cairo.jl
```diff
@@ -438,6 +438,13 @@ function stroke_preserve(ctx::CairoContext)
     restore(ctx)
 end
 
+function paint_with_alpha(ctx::CairoContext, a = 0.5)
```
nalimilan: How useful is the `a=0.5` default? Maybe it would be better not to provide any default?

src/Cairo.jl
nalimilan: And you should be able to use only two lines, instead of four.

nalimilan:
Thanks!

Regarding tests, I think you can take inspiration from the uses of `paint` in `test/shape_functions.jl`. You could imagine changing one call to `paint` into `paint_with_alpha`.

cormullion:
@nalimilan Thanks so much for the suggestions, I'm definitely outside my comfort zone atm. Will add some more commits tomorrow. 

The 0.5 default is just me bringing habits over from using Mathematica. I don't see a good reason for it, now you mention it... :)

tkelman:
This looks okay to me, are you happy with it now @cormullion?

cormullion:
@tkelman Aye, it looks fine and works OK.

Issue #133, opened 2016-04-26
Building Cairo.jl for AWS Lambda


samoconnor:
Cairo.jl does not build cleanly for AWS Lambda.

The AWS Lambda sandbox runs on [Amazon Linux](https://aws.amazon.com/amazon-linux-ami/).
An [equivalent EC2 image](https://github.com/samoconnor/AWSLambda.jl/blob/master/src/AWSLambda.jl#L1187) is used to build binaries that will run on AWS Lambda.

There are a number of problems with the way Cairo.jl's BinDeps configuration works at present:
1. It tries to use `yum` to install dependencies. This "works" on the EC2 build box, but creates binaries that depend on libraries that do not exists in the Lambda sandbox. I have worked around this by disabling `yum` on EC2 `chmod 000 /usr/bin/yum`.
2. It sees that `libfreetype` and `libfontconfig` are installed, and therefore does not try to build them from source, but this causes the `pango` build to fail later because `pango` requires more recent versions. I've worked around this by modifying the `aliases =` entries to specify specific version numbers.
3. It seems that `pango` requires `libharfbuzz`, but there is no rule to build this in `build.jl`. I've added an `Autotools` rule for this.
4. `libgobject` is in the same "group" as all the other deps. This results in an empty `# Load dependencies` section in `deps.jl` because `libgobject` is satisfied by a system lib and the others in the group are satisfied by `Autotools`.
5. There seem to be a number of deps that are labeled `runtime = false` that are in fact needed at runtime. e.g. `libpng`, `libpixman`, `libffi`, `libfontconfig`
6. `fontconfig` needs the option `--enable-libxml2`, otherwise it requires expat to be installed.

After some experimentation (see patch below), I've managed to get the build working with the following output in `deps.jl`:

``` julia
@checked_lib _jl_libgobject "/usr/lib64/libgobject-2.0.so"
@checked_lib _jl_libpango "/var/task/julia/v0.4/Cairo/deps/usr/lib/libpango-1.0.so"
@checked_lib _jl_libpangocairo "/var/task/julia/v0.4/Cairo/deps/usr/lib/libpangocairo-1.0.so"
@checked_lib _jl_libcairo "/var/task/julia/v0.4/Cairo/deps/usr/lib/libcairo.so" 
```

[...]

PR #134, opened 2016-04-28
Building Cairo.jl for AWS Lambda

samoconnor:
See #133 

Amazon Linux on the Lambda sandbox meets some of Cairo.jl's dependancies with system libraries.
There is no way to install new system libraries (i.e. no root access and no yum). 
`freetype` and `fontconfig` are installed as system libraries but the versions are too old for `pango`.

This PR allows Cairo.jl to use the system versions of `libgobject`, `libffi`, `libpng` and `libpixman`, while everything else is built form source.

The commits make one change at a time:
- Remove runtime = false for libraries that are required at runtime (see ldd output in #133).
- remove gettext (nothing depends on it)
- Add harfbuzz (required by pango)
- Update libfontconfig version. (and don't try to build libfontconfig against expat).
- add option JULIA_BINDEPS_DISABLE_SYSTEM_PACKAGE_MANAGERS
- add option JULIA_BINDEPS_IGNORE_SYSTEM_FONT_LIBS
- split dependancies into two groups (otherwise no `@checked_lib` statements are generated in `deps.jl`).

The resulting `deps.jl` contains this:

``` julia
@checked_lib _jl_libgobject "/usr/lib64/libgobject-2.0.so"
@checked_lib _jl_libpango "/var/task/julia/v0.4/Cairo/deps/usr/lib/libpango-1.0.so"
@checked_lib _jl_libpangocairo "/var/task/julia/v0.4/Cairo/deps/usr/lib/libpangocairo-1.0.so"
@checked_lib _jl_libcairo "/var/task/julia/v0.4/Cairo/deps/usr/lib/libcairo.so"
```

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -11,7 +11,7 @@ ignore_system_font_libs(n, h) = !(haskey(ENV, "JULIA_BINDEPS_IGNORE_SYSTEM_FONT_
 
 deps = [
     libffi = library_dependency("ffi", aliases = ["libffi"], group = group1)
-    gobject = library_dependency("gobject", aliases = ["libgobject-2.0-0", "libgobject-2.0", "libgobject-2_0-0", "libgobject-2.0.so.0"], depends=[libffi], group = group1)
+    gobject = library_dependency("gobject", aliases = ["libgobject-2.0.so.0", "libgobject-2.0-0", "libgobject-2.0", "libgobject-2_0-0"], depends=[libffi], group = group1)
     zlib = library_dependency("zlib", aliases = ["libzlib","zlib1"], os = :Windows, group = group1)
     libpng = library_dependency("png", aliases = ["libpng","libpng-1.5.14","libpng15","libpng12.so.0"], group = group1)
     pixman = library_dependency("pixman", aliases = ["libpixman","libpixman-1","libpixman-1-0","libpixman-1.0"], depends = [libpng], group = group1)


tkelman:
is harfbuzz needed everywhere? will need to add package manager sources for it if so, and provide a windows binary source.

the validate hook should just check for required versions or symbols rather than paths and environment variables.

lobingera:
@tkelman; i'm not an pango expert, but their web-page claims:

> Pango is designed to be modular; the core Pango layout engine can be used with different font backends. There are three basic backends, with multiple options for rendering with each.

>  - Client side fonts using the FreeType and fontconfig libraries, using HarfBuzz for complex-text handling. Rendering can be with with Cairo or Xft libraries, or directly to an in-memory buffer with no additional libraries.
>  - Native fonts on Microsoft Windows using Uniscribe for complex-text handling. Rendering can be done via Cairo or directly using the native Win32 API.
>  - Native fonts on MacOS X using CoreText for complex-text handling, rendering via Cairo. 

so there seems to be a propability to setup a native system without harfbuzz

tkelman:
Not sure. We get all these binaries from winrpm, all the windows-specific BuildProcess lines can likely be deleted since I don't think they've been used in 2+ years. I turned on AppVeyor since it was easy and the package currently works on Windows, better to keep it that way automatically while tweaking BinDeps. The opensuse build service that winrpm downloads from has the cross-compiled libraries including harfbuzz, so that just gets downloaded and seems to work fine. May need to adjust the winrpm `provides` lines if harfbuzz is going to be added as a separate `library_dependency` is all.

samoconnor:
> the validate hook should just check for required versions or symbols rather than paths and environment variables.

But how would we know what version of `fontconfig` to check for?

If `fontconfig` version x is installed on my system already, then the `pango` that is installed by `yum install pango` will probably work fine with `fontconfig` version x. However, the `pango` version installed by Autoconf might require `fontconfig` version y. There's no way that all the interdependent versions can be kept in sync all the time across platforms and distributions.

In this case I need a way to say "BinDeps should ignore system libraries A and B".

tkelman:
The library group functionality is supposed to be preventing that mixed-source scenario of trying to use some libraries from the system and some from source. If it's not doing so that sounds like it may be a BinDeps bug, or a misusage here.

tkelman:
`JULIA_BINDEPS_DISABLE_SYSTEM_PACKAGE_MANAGERS` seems reasonable to me (though it's getting awfully long, maybe drop the `SYSTEM` part to shorten it up a bit?) but would be better-suited to implement in BinDeps rather than in every package that uses it.

deps/build.jl
```diff
@@ -69,19 +73,20 @@ provides(Yum,
...
 const png_version = "1.5.14"
 
 provides(Sources,
     @compat Dict(
         URI("http://www.cairographics.org/releases/pixman-0.28.2.tar.gz") => pixman,
         URI("http://www.cairographics.org/releases/cairo-1.12.16.tar.xz") => cairo,
         URI("http://download.savannah.gnu.org/releases/freetype/freetype-2.4.11.tar.gz") => freetype,
-        URI("http://www.freedesktop.org/software/fontconfig/release/fontconfig-2.10.2.tar.gz") => fontconfig,
-        URI("http://ftp.gnu.org/pub/gnu/gettext/gettext-0.18.2.tar.gz") => gettext,
+        URI("https://www.freedesktop.org/software/harfbuzz/release/harfbuzz-1.2.6.tar.bz2") => harfbuzz,
```
nalimilan: Use HTTPS everywhere possible, or nowhere.

deps/build.jl
```diff
@@ -95,12 +100,12 @@ provides(BuildProcess,
     @compat Dict(
         Autotools(libtarget = "pixman/libpixman-1.la", installed_libname = xx("libpixman-1-0.","libpixman-1.","libpixman-1.0.")*BinDeps.shlib_ext) => pixman,
         Autotools(libtarget = xx("objs/.libs/libfreetype.la","libfreetype.la")) => freetype,
-        Autotools(libtarget = "src/libfontconfig.la") => fontconfig,
+        Autotools(libtarget = "src/libharfbuzz.la") => harfbuzz,
+        Autotools(libtarget = "src/libfontconfig.la", configure_options = ["--enable-libxml2"]) => fontconfig,
```
nalimilan: I guess we need to add libxml2 too?

deps/build.jl
```diff
@@ -58,8 +63,7 @@ provides(AptGet,
         "libpango1.0-0" => [pango,pangocairo],
         "libglib2.0-0" => gobject,
         "libpng12-0" => libpng,
-        "libpixman-1-0" => pixman,
-        "gettext" => gettext
```
nalimilan: I also wanted to remove this some time ago, but it'ss actually needed by GLib. Though I'm not sure anybody has already tested building everything from source. See https://github.com/JuliaLang/Cairo.jl/issues/36.

deps/build.jl
```diff
@@ -11,7 +11,7 @@ ignore_system_font_libs(n, h) = !(haskey(ENV, "JULIA_BINDEPS_IGNORE_SYSTEM_FONT_
 
 deps = [
     libffi = library_dependency("ffi", aliases = ["libffi"], group = group1)
-    gobject = library_dependency("gobject", aliases = ["libgobject-2.0-0", "libgobject-2.0", "libgobject-2_0-0", "libgobject-2.0.so.0"], depends=[libffi], group = group1)
+    gobject = library_dependency("gobject", aliases = ["libgobject-2.0.so.0", "libgobject-2.0-0", "libgobject-2.0", "libgobject-2_0-0"], depends=[libffi], group = group1)
```
nalimilan: Does this make any difference? If so, the commit message should give it.

deps/build.jl
samoconnor: > [gettext is] actually needed by GLib

I don't know the details, but I suspect that it depends on how GLib is built. Amazon Linux has GLib but not gettext.

nalimilan: Probably, but I suspect you need it if you want GTK+ stock messages to be translated. And not everybody speaks English out there.

deps/build.jl
samoconnor: The commit message is "put more specific gobject alias first". This ensures that the more specific pattern is tried first. In this case it makes a difference because the build box has a symlink "libgobject-2.0.so -> libgobject-2.0.so.0" but the deployment box does not  (it only has "libgobject-2.0.so.0"). Putting "libgobject-2.0.so.0" first ensures that this is what ends up in the generated code.

giordano:
Probably also this one is not necessary after #292

davidanthoff:
Yeah, I think this PR should be closed. If it is still needed it would need a thorough update, but most likely it is just no longer relevant.

PR #136, opened 2016-05-15
Compat update for String rename

StefanKarpinski:


yuyichao:
This changes the return type and is potentially breaking.

timholy:
Doesn't it just change spelling? On julia-0.4:
```jl
julia> using Compat; using Compat.String

julia> String
Union{ASCIIString,UTF8String}

julia> ByteString
Union{ASCIIString,UTF8String}
```

yuyichao:
Ah, yeah, this one is fine, I think i might have misread the `ByteString` as `UTF8String`

yuyichao:
AppVeyor failure seems to be unrelated. This requires `Compat 0.7.15`

timholy:
Yes, the REQUIRE file should be updated.

timholy:
That same AppVeyor error appears at https://github.com/JuliaIO/HDF5.jl/pull/302

PR #139, opened 2016-06-17
Allow RGB24/ARGB32 eltype in CairoSurface; fix depwarns

timholy:
This overdue change simply allows one to create a `CairoSurface` in terms of a `Matrix{T}` where  `T` is `RGB24` or `ARGB32` (from ColorTypes.jl). This allows Cairo to infer the (Cairo) format without having to mess with extra arguments.

Also fixes a number of deprecations on julia-0.5.

### DIFF
--- a/src/Cairo.jl
+++ b/src/Cairo.jl
@@ -95,18 +95,17 @@ else
     get_stream_callback(T) = cfunction(write_to_stream_callback, Int32, (Ref{T}, Ptr{UInt8}, UInt32))
 end
 
-type CairoSurface <: GraphicsDevice
+type CairoSurface{T<:@compat(Union{UInt32,RGB24,ARGB32})} <: GraphicsDevice
     ptr::Ptr{Void}
     width::Float64
     height::Float64
-    data::Array{UInt32,2}
+    data::Matrix{T}
 
     function CairoSurface(ptr::Ptr{Void}, w, h)
         self = new(ptr, w, h)
         finalizer(self, destroy)
         self
     end
-
     function CairoSurface(ptr::Ptr{Void}, w, h, data)
         self = new(ptr, w, h, data)
         finalizer(self, destroy)
@@ -117,10 +116,10 @@ type CairoSurface <: GraphicsDevice
         #finalizer(self, destroy)
         self
     end
-
-
-
 end
+
+CairoSurface(ptr, w, h) = CairoSurface{UInt32}(ptr, w, h)
+CairoSurface{T}(ptr, w, h, data::Matrix{T}) = CairoSurface{T}(ptr, w, h, data)
 
 width(surface::CairoSurface) = surface.width
 height(surface::CairoSurface) = surface.height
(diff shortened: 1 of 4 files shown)


timholy:
I also noticed that the tests are fairly anemic. In an attempt to help redress this, I am now running all the `sample.s/*.jl` files (contributed by @lobingera) as part of the tests. I'm not comparing the outputs to any kind of standard, but at least this is a step in the right direction.

lobingera:
@timholy please open an issue for testing.

lobingera:
It's not that I'm too lazy, i don't have full Internet Access in my holiday resort ...

timholy:
Done, #140. Hope you're having a nice holiday!

src/Cairo.jl
```diff
@@ -311,10 +323,10 @@ end
 
 function write_to_png(surface::CairoSurface, filename::AbstractString)
     ccall((:cairo_surface_write_to_png,_jl_libcairo), Void,
-          (Ptr{UInt8},Ptr{UInt8}), surface.ptr, bytestring(filename))
+          (Ptr{UInt8},Ptr{UInt8}), surface.ptr, @compat(String(filename)))
 end
 
-writemime(io::IO, ::MIME"image/png", surface::CairoSurface) =
+@compat Base.show(io::IO, ::MIME"image/png", surface::CairoSurface) =
```
tkelman: Compat minimum version for this?

timholy: Hmm, yes you're right (as always---thanks).

Should I amend METADATA for the recent release, too?

tkelman: We can amend metadata if you'd like to be especially thorough, sure. Minimum versions not being recent enough hasn't caused too many issues yet (adding an upper bound on Compat would be a bit strange), but I think it's worth generally keeping in mind.

tkelman: Looks like AppVeyor is now running on a JuliaGraphics account instead of my personal account, great. Could you (or someone else? @simondanisch maybe?) grant me collaborator access to that appveyor account? It would be good to enable a daily scheduled build here to catch potential WinRPM issues triggered by upstream changes. Also the readme appveyor badge needs updating.

timholy: I didn't realize you had it set up on your account---I thought it wasn't running PRs, because it didn't show up when I pushed things. I basically created a new account called JuliaGraphics, but it's emailing me. If you'd like to be the main administrator of this account, by all means please do so. I'll see if I can figure out how to set that up.

tkelman: Something was going wrong with the webhook, not sure what. AppVeyor doesn't always handle repository moves all that smoothly. Collaborator or admin access would be fine - I think you can disable the emails somewhere in their settings, but it doesn't seem to obey disabling email notifications from the yml as well as travis does.

timholy: Should be done. Feel free to change *anything* in how that account is administered, delete the account and start over, etc.

Issue #141, opened 2016-06-22
CairoContext from pointer constructor fails


Matt5sean3:
The `CairoContext` constructor that takes a `Ptr{Void}` as its argument fails. 

The error message is shown here
```
ERROR: LoadError: MethodError: Cannot `convert` an object of type Ptr{Void} to an object of type Cairo.CairoSurface{T<:Union{ColorTypes.ARGB32,ColorTypes.RGB24,UInt32}}
This may have arisen from a call to the constructor Cairo.CairoSurface{T<:Union{ColorTypes.ARGB32,ColorTypes.RGB24,UInt32}}(...),
since type constructors fall back to convert methods.
 in Cairo.CairoContext(::Ptr{Void}) at /home/matthew/.julia/v0.5/Cairo/src/Cairo.jl:367
 in include_from_node1(::String) at ./loading.jl:426
 in process_options(::Base.JLOptions) at ./client.jl:266
 in _start() at ./client.jl:322
while loading /home/matthew/test.jl, in expression starting on line 33
```

The cause of the failure seems to be that `CairoSurface` lacks a constructor that takes only a `Ptr{Void}` as its argument but the `CairoContext` constructor attempts to use a `CairoSurface` constructor of that form.

I've tested this with both the master version of the repository and the current release. I've tested using both julia 0.4.5 and pre-release 0.5. The issue appears in all of those cases.

The issue can be observed using the code below.

```julia
using Gtk
using Cairo

da = Gtk.@GtkDrawingArea(ccall((:gtk_drawing_area_new, Gtk.libgtk), Ptr{Gtk.GObject}, ()))

setproperty!(da, "width-request", 300)
setproperty!(da, "height-request", 300)

win = @GtkWindow(da, "Test Window")

showall(win)

w = Gtk.width(da)
h = Gtk.height(da)
surf_ptr = 
  ccall(
    (:gdk_window_create_similar_surface, Gtk.libgdk), 
    Ptr{Void}, 
    (Ptr{Void}, Gtk.GEnum, Cint, Cint),
    Gtk.gdk_window(da),
    Cairo.CONTENT_COLOR_ALPHA,
    w, 
    h)

ctx_ptr = ccall(
    (:cairo_create, Cairo._jl_libcairo), 
    Ptr{Void}, 
    (Ptr{Void}, ), 
    surf_ptr)

ctx = CairoContext(ctx_ptr)
```

Matt5sean3:
[...]

PR #142, opened 2016-06-22
Support creating CairoContext and CairoSurface objects from Ptr{Void}

Matt5sean3:
This is an attempt at resolving issue #141.

* Primary piece for solving #141 is line 126 of the new Cairo.jl.
* When only the pointer is provided the reference count is incremented which prevents deletion but provides the behavior that appears to have been desired.
* A test is added so that future changes will cause a failed test if they break this functionality for `CairoContext` objects.


Matt5sean3:
`finalize` is breaking compatibility on julia 0.3 and Cairo's reference counting on travis-ci looks to be downright broken so I'm removing the test for reference counting.

timholy:
Thanks for the fix!

PR #155, opened 2016-08-22
read png from stream

lobingera:
This deals as preparation of https://github.com/dcjones/Compose.jl/pull/141.
What's missing:

- testing outside linux+0.5
- sample description
- (maybe) some wording


src/Cairo.jl
```diff
@@ -100,8 +100,12 @@ end
 
 function read_from_stream_callback(s::IO, buf::Ptr{UInt8}, len::UInt32)
     # wrap the provided buf into a julia Array
-    b1 = unsafe_wrap(Array{UInt8,1},buf,len)
 
+    if VERSION < v"0.5.0-rc"
+        b1 = pointer_to_array(buf,len); #pre-v0.5
+    else
+        b1 = unsafe_wrap(Array{UInt8,1},buf,len);
```
tkelman: this should be in recent Compat

lobingera: I've seen this after committing ... but i had strange testing results (OSX succ, linux fail) with 0.5, so i need another run anyway.

lobingera: @tkelman 

Maybe i'm wrong, but 
```
using Compat
...
b1 = unsafe_wrap(Array{UInt8,1},buf,len);
```
gives me
```
ERROR: LoadError: MethodError: `unsafe_wrap` has no method matching unsafe_wrap(::Type{Array{UInt8,1}}, ::Ptr{UInt8}, ::UInt32)
Closest candidates are:
  unsafe_wrap(::Type{ByteString}, ::Union{Ptr{Int8},Ptr{UInt8}}, ::Any)
  unsafe_wrap(::Type{ByteString}, ::Union{Ptr{Int8},Ptr{UInt8}}, ::Any, ::Bool)
  unsafe_wrap(::Type{Array{T,N}}, ::Ptr{T}, ::Any)
  ...
 in read_from_stream_callback at /home/lobi/juliarepo/png_stream.jl:11
while loading /home/lobi/juliarepo/png_stream.jl, in expression starting on line 58
```
in a v0.4.5 with Compat 0.8.8

tkelman: looks like https://github.com/JuliaLang/Compat.jl/blob/1a19f3ed0fb4062a18aabdd0d2adfb84ff794980/src/Compat.jl#L1393 might need a few more parametrized methods?

src/Cairo.jl
```diff
@@ -118,7 +118,12 @@ function read_from_stream_callback(s::IO, buf::Ptr{UInt8}, len::UInt32)
     r 
 end
 
-get_readstream_callback(T) = cfunction(read_from_stream_callback, Int32, (Ref{T}, Ptr{UInt8}, UInt32))
+
+if VERSION < v"0.4-"
```
tkelman: we should just drop 0.3 support

src/Cairo.jl
lobingera: So i file this as Compat issue?

lobingera:
The error on travis nightly linux seems to be an issue with GC and a cairo destroy function. It's hard to reproduce, as it seems to be a memory problem not in the new code, but triggered by the sequence of testing.

lobingera:
If we exclude - like Tony proposed - 0.3 this is close to ready.

src/Cairo.jl
```diff
@@ -97,6 +97,36 @@ else
...
 end
 
+
+function read_from_stream_callback(s::IO, buf::Ptr{UInt8}, len::UInt32)
+    # wrap the provided buf into a julia Array
+
+    if VERSION < v"0.5.0-rc"
+        b1 = pointer_to_array(buf,len); #pre-v0.5
+    else
+        b1 = unsafe_wrap(Array{UInt8,1},buf,len);
```
timholy: This version is supported by Compat, so I think you don't need the VERSION check. https://github.com/JuliaLang/Compat.jl#renamed-functions

src/Cairo.jl
```diff
@@ -335,6 +365,18 @@ end
 @compat show(io::IO, ::MIME"image/png", surface::CairoSurface) =
    write_to_png(surface, io)
 
+function read_from_png_stream{T<:IO}(stream::T)
```
timholy: Does this really have to be a separately-named function? Why not just make it a method of `read_from_png`?

test/runtests.jl
```diff
@@ -44,6 +44,7 @@ test_pattern_get_surface()
 pth = joinpath(dirname(dirname(@__FILE__)), "samples")
 fls = filter(str->endswith(str,".jl"), readdir(pth))
 for fl in fls
+    println(fl)
```
timholy: left over from debugging, or intended?

lobingera: just added for debugging

src/Cairo.jl
lobingera: agreed

src/Cairo.jl
lobingera: (see discussion in https://github.com/JuliaGraphics/Cairo.jl/pull/155/commits/1934e05269372117b690d2e1ed43fd7fe5b83908)

lobingera:
as i wondered in https://github.com/JuliaGraphics/Cairo.jl/pull/155#issuecomment-242906598
and testing without sample_copy_path suceeds (locally and on travis/appveyor) i try see what's wrong with memory management in copypath

src/Cairo.jl
timholy: I think it should just be `unsafe_wrap(Array, buf, len)`.

test/runtests.jl
timholy: Carry on, then.

lobingera: i'm a little bit stuck. I get **very** strange signal(6) and signal(11) around line 15 of sample_imagestream which the input string (inlined png). From that i worked backwards and somehow got the impression, that copy_path is creating the problems, as initially i got (always) the signal(6) from a destroy(path) call (in libcairo) and that's the only where a path object is created. 
Now i got:
```
lobi@orange4:~/juliarepo$ ../julia05/julia 
               _
   _       _ _(_)_     |  A fresh approach to technical computing
  (_)     | (_) (_)    |  Documentation: http://docs.julialang.org
   _ _   _| |_  __ _   |  Type "?help" for help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 0.5.0-rc3+0 (2016-08-22 23:43 UTC)
 _/ |\__'_|_|_|\__'_|  |  Commit e6f843b* (4 days old release-0.5)
|__/                   |  x86_64-linux-gnu

julia> Pkg.test("Cairo")
INFO: Testing Cairo
sample_arc.jl
sample_arc_negative.jl
sample_clip_image.jl
sample_clip.jl
sample_copy_path.jl
sample_curve_rectangle.jl
sample_curve_to.jl
sample_dash.jl
sample_fill_and_stroke2.jl
sample_fill_style.jl
sample_gradient.jl
sample_image.jl
sample_imagepattern.jl
sample_imagestream.jl
*** Error in `/home/lobi/julia05/usr/bin/julia': double free or corruption (out): 0x0000000002d9e1f0 ***

signal (6): Aborted
while loading /home/lobi/.julia/v0.5/Cairo/samples/sample_imagestream.jl, in expression starting on line 15
raise at /build/eglibc-3GlaMS/eglibc-2.19/signal/../nptl/sysdeps/unix/sysv/linux/raise.c:56
abort at /build/eglibc-3GlaMS/eglibc-2.19/stdlib/abort.c:89
__libc_message at /build/eglibc-3GlaMS/eglibc-2.19/libio/../sysdeps/posix/libc_fatal.c:175
malloc_printerr at /build/eglibc-3GlaMS/eglibc-2.19/malloc/malloc.c:4996 [inlined]
_int_free at /build/eglibc-3GlaMS/eglibc-2.19/malloc/malloc.c:3840
jl_gc_free_array at /home/lobi/julia05/src/gc.c:744 [inlined]
sweep_malloced_arrays at /home/lobi/julia05/src/gc.c:765 [inlined]
gc_sweep_other at /home/lobi/julia05/src/gc.c:1032 [inlined]
_jl_gc_collect at /home/lobi/julia05/src/gc.c:1792 [inlined]
```
and (within the same session)
```
julia> Pkg.test("Cairo")
INFO: Testing Cairo
sample_arc.jl
sample_arc_negative.jl
sample_clip_image.jl
sample_clip.jl
sample_copy_path.jl
sample_curve_rectangle.jl
sample_curve_to.jl
sample_dash.jl
sample_fill_and_stroke2.jl
sample_fill_style.jl
sample_gradient.jl
sample_image.jl
sample_imagepattern.jl
sample_imagestream.jl
julia: /home/lobi/julia05/deps/srccache/llvm-3.7.1/lib/IR/Value.cpp:175: void llvm::Value::setValueName(llvm::ValueName*): Assertion `HasName == Ctx.pImpl->ValueNames.count(this) && "HasName bit out of sync!"' failed.

signal (6): Aborted
while loading /home/lobi/.julia/v0.5/Cairo/samples/sample_imagestream.jl, in expression starting on line 15
raise at /build/eglibc-3GlaMS/eglibc-2.19/signal/../nptl/sysdeps/unix/sysv/linux/raise.c:56
abort at /build/eglibc-3GlaMS/eglibc-2.19/stdlib/abort.c:89
__assert_fail_base at /build/eglibc-3GlaMS/eglibc-2.19/assert/assert.c:92
__assert_fail at /build/eglibc-3GlaMS/eglibc-2.19/assert/assert.c:101
_ZN4llvm5Value12setValueNameEPNS_14StringMapEntryIPS0_EE at /home/lobi/julia05/usr/bin/../lib/libLLVM-3.7.1.so (unknown line)
_ZN4llvm5Value11setNameImplERKNS_5TwineE at /home/lobi/julia05/usr/bin/../lib/libLLVM-3.7.1.so (unknown line)
_ZN4llvm5Value7setNameERKNS_5TwineE at /home/lobi/julia05/usr/bin/../lib/libLLVM-3.7.1.so (unknown line)
_ZN4llvm8FunctionC2EPNS_12FunctionTypeENS_11GlobalValue12LinkageTypesERKNS_5TwineEPNS_6ModuleE at /home/lobi/julia05/usr/bin/../lib/libLLVM-3.7.1.so (unknown line)
Create at /home/lobi/julia05/usr/include/llvm/IR/Function.h:123 [inlined]
function_proto at /home/lobi/julia05/src/jitlayers.h:95
emit_jlcall at /home/lobi/julia05/src/codegen.cpp:2592
emit_call_function_object at /home/lobi/julia05/src/codegen.cpp:2662
emit_invoke at /home/lobi/julia05/src/codegen.cpp:2693 [inlined]
emit_expr at /home/lobi/julia05/src/codegen.cpp:3165
emit_function at /home/lobi/julia05/src/codegen.cpp:4694
jl_compile_linfo at /home/lobi/julia05/src/codegen.cpp:810
jl_compile_for_dispatch at /home/lobi/julia05/src/gf.c:1311
...
```

i need a better idea, what happens (are working debugger...)

timholy: Rather than an IO buffer, try it first with a different kind of stream instead:
```jl
img = open(filename) do io
    read_from_png(io)
end
```
It's quite possible that the Cairo library won't support julia's IOBuffer type, which is not a stream in the Unix sense (C libraries are oh-so-picky). Note:
```jl
julia> supertype(typeof(STDIN))
Base.LibuvStream

julia> supertype(supertype(typeof(STDIN)))
IO

julia> supertype(IOBuffer)
IO
```

lobingera: I did choose the IOBuffer as interface, libcairo actually only needs a function call https://www.cairographics.org/manual/cairo-PNG-Support.html#cairo-image-surface-create-from-png-stream that copies sequentially from input to an internal buffer. As my initial target was "read png from string" IOBuffer seemed a good match.

.travis.yml
```diff
@@ -3,7 +3,6 @@ os:
     - linux
     - osx
 julia:
-    - 0.3
```
tkelman: don't remove testing for it unless you also change the minimum version in REQUIRE. that's what actually matters to thr package manager

lobingera:
This was rather a Test, there is an issue for reducing #159

lobingera:
for those following this:
it's now a version within a single commit (re-based), but runs only > v0.3

lobingera:
So.

* ready and tested for 0.4/0.5/0.6. 
* rebased from master
* read_from_png dispatched on stream
* example to read png from a string (inline)

@JuliaGraphics/collaborators: Please pull (i still don't have write access...)

src/Cairo.jl
timholy: That link doesn't work anymore, can you explain? Also, you don't need the semicolons. (They don't hurt anything, of course, but they are unusual.)

src/Cairo.jl
```diff
@@ -89,6 +89,33 @@ end
...
+        b1 = pointer_to_array(buf,len); #pre-v0.5
+    else
+        b1 = unsafe_wrap(Array{UInt8,1},buf,len);
+    end
+    # read from stream
+    nb = readbytes!(s,b1,len)
+
+    # provide a return status
+    if nb == len
+        r = @compat(Int32(0)) # CAIRO_STATUS_SUCCESS
```
timholy: Given that both of those constants are defined (see `constants.jl`), why not use them directly?

src/Cairo.jl
```diff
@@ -89,6 +89,33 @@ end
...
+    else
+        r = @compat(Int32(10)) # CAIRO_STATUS_READ_ERROR
+    end
+    r
+end
+
+
+get_readstream_callback(T) = cfunction(read_from_stream_callback, Int32, (Ref{T}, Ptr{UInt8}, UInt32))
+
+
```
timholy: There's more blank lines here (and perhaps above) than really necessary.

src/Cairo.jl
lobingera: agreed. I was distracted, because the write_to_stream callbacks also use hardcoded values.

src/Cairo.jl
lobingera: the usafe_wrap discussion is in the discussion to the first outdated diff (tony, 23aug)
and there i didn't understand your comment on  "it should just be unsafe_wrap(Array, buf, len)" Don't we need type here?

timholy: Don't need it:

```jl
julia> a = rand(Float32, 5)
5-element Array{Float32,1}:
 0.304382
 0.980674
 0.91572 
 0.626573
 0.960012

julia> p = pointer(a)
Ptr{Float32} @0x00007f207e3fd390

julia> unsafe_wrap(Array, p, 5)
5-element Array{Float32,1}:
 0.304382
 0.980674
 0.91572 
 0.626573
 0.960012
```

timholy: But it would be good to at least file an issue with Compat. (Or just add the missing methods as a PR.)

timholy:
Thanks!

Issue #156, opened 2016-09-13
linux julia-0.5 segfault on sample_imagepattern


lobingera:
https://github.com/JuliaGraphics/Cairo.jl/issues/146

And it goes something like this:
```
julia> versioninfo()
Julia Version 0.5.0-rc4+0
Commit 9c76c3e* (2016-09-09 01:43 UTC)
DEBUG build
Platform Info:
  System: Linux (x86_64-linux-gnu)
  CPU: Intel(R) Core(TM) i3-2120 CPU @ 3.30GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Sandybridge)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.7.1 (ORCJIT, sandybridge)
```
i run repeatedly Pkg.test("Cairo") and get signal 6, signal 11 and sometimes julia freezes, then crtl-C brings one of the signals. The top-level call is sometimes in libc-signal, sometimes in uv_ sometimes in codegen, sometimes in jit sometimes directly in flisp.c.

A longer xterm trace will be found in a gist.

lobingera:
https://gist.github.com/lobingera/249a459ef2f1b9801033596fc6f7369c#file-xterm-log-orange4-2016-09-13-21-12-20-2076

lobingera:
following Yichao Yu's analysis in https://groups.google.com/forum/#!topic/julia-users/b3zdcMhZSVg the orginal error is in Cairo.jl line 625
```
c_data = @compat unsafe_wrap(Array,c.data,(Int(c.num_data*2),1),true)
```
which 'owns' the copied memory which later can lead to double free.
PR #157

Issue #146, opened 2016-06-24
OSX julia-0.5 segfault on sample_imagepattern


timholy:
I can't reproduce this on linux, so I can't fix it, but see e.g., https://travis-ci.org/JuliaGraphics/Cairo.jl/builds/139925587, and click on the OSX nightly run. It may not be 100% reliable, but it's shown up on the majority of recent travis builds.

lobingera:
just for more context, message copied from travis log
```
julia(6265,0x7fff77df3310) malloc: *** error for object 0x7fe54d281f60: pointer being freed was not allocated

*** set a breakpoint in malloc_error_break to debug

signal (6): Abort trap: 6

while loading /Users/travis/.julia/v0.5/Cairo/samples/sample_imagepattern.jl, in expression starting on line 29

__pthread_kill at /usr/lib/system/libsystem_kernel.dylib (unknown line)

Allocations: 3596950 (Pool: 3595463; Big: 1487); GC: 5
```
@timholy Do you have experience to track down, what instance is messaging what here? (it's julia itself, seeing a memory error?)
and btw: https://travis-ci.org/JuliaGraphics/Cairo.jl/builds/139835595 passed

lobingera:
i tried to recreate with a dummy package running only sample_imagepattern and running Pkg.test() on my Cairo.jl fork.

https://travis-ci.org/lobingera/L1Dummy/builds/140205864
https://travis-ci.org/JuliaGraphics/Cairo.jl/builds/140209942
Both passes.

tkelman:
saw a "double free or corruption" on pkgeval (ubuntu 14.04, 64 bit) at this same place here: https://github.com/JuliaLang/julia/pull/18276#issuecomment-245477100

lobingera:
I mentioned this in my comment https://github.com/JuliaLang/julia/pull/18276#issuecomment-244896049 and https://github.com/JuliaGraphics/Cairo.jl/pull/155#issuecomment-242906598.

The error "double free or corruption" first only showed up on travis and only on linux, later i could see it with my local linux install also. The point when it's happening seemed to be triggered by GC when julia calls destroy on elements allocated by libcairo.

PR #157, opened 2016-09-14
copy path output should not be owned

lobingera:
#146,#156 
the output pointer of copy path should not be owned by julia as libcairo (via glib) has own memory management.

### DIFF
--- a/src/Cairo.jl
+++ b/src/Cairo.jl
@@ -622,7 +622,7 @@ function convert_cairo_path_data(p::CairoPath)
     # define here by Float64 (most data is) and reinterpret in the header.
 
     path_data = CairoPathEntry[]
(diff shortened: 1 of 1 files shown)


yuyichao:
LGTM. AppVeyor has a download error.

P.S. I believe the `false` is the default.

lobingera:
@yuyichao Can this be merged, even with the appveyor (download) error? Or do i need to push 'something' to restart? And if the first, can you do it?

tkelman:
(ps should drop 0.3 support before tagging this, then do so with `Pkg.tag("Cairo", :minor)`)

tkelman:
Actually given this is a segfault fix, I guess we can tag it for 0.3 too and make an exception.

PR #163, opened 2016-10-02
WIP: more testing

lobingera:


test/test_painting.jl
```diff
@@ -0,0 +1,199 @@
+# some collection of some painting plus support functions
+
+"""
+function ngray(base::Int64,digits::Int64,value::Int64)
+Convert a value to a graycode with the given base and digits
+"""
+function ngray(base::Int64,digits::Int64,value::Int64)
+	baseN = zeros(Int64,digits)
```
tkelman: be consistent about tabs vs spaces?

lobingera: agreed. There is a reason, why it's called work in progress...

lobingera:
@timholy, @tkelman Mac OSX 10.9 tests failing on all travis with warnings about 10.9 and failing by installing gobject-introspection. Should we raise the Mac OSX version anyway?

tkelman:
Ah yeah, homebrew no longer supports that version. Ref https://blog.travis-ci.com/2016-09-15-new-default-osx-image-coming/, hopefully we're less than a day from Travis changing the default image, but guess you can change it now if you'd like.

lobingera:
@tkelman For the travis OSX: i'll just wait for an update on their side. 
@timholy : How can we (or you) establish a reference point for coveralls? Only changes unknown is reported here.

REQUIRE
```diff
@@ -3,5 +3,6 @@ Compat 0.8.0
 Colors
 BinDeps 0.3.21
 Graphics 0.1
+BaseTestNext
```
tkelman: this should go in `test/REQUIRE`, not `REQUIRE` since it's only needed for tests

lobingera: I didn't We specialize for testing only, i'll Change in the next commit

test/REQUIRE
```diff
@@ -0,0 +1,8 @@
+julia 0.3
```
tkelman: don't need to duplicate, only list things that are stricter or not included in the top level `REQUIRE`

lobingera: Thanks for the pointer. To simplify this, where is the test/REQUIRE documented?

tkelman: Hm, good question. It might not be! If we can't find it, or any open issues about it, then we should open a doc to-do issue to add some.

timholy:
Yay for more testing. These looks like mostly a long string of Cairo calls, so while I haven't looked over the code in detail I can't imagine having big concerns. I'd say merge when you're happy with it.

lobingera:
@timholy Recently i'm not so happy with the tests (incl. build) failing for 0.6 ... otherwise i would have rebased yesterday ...

lobingera:
superceded by #167

PR #167, opened 2017-01-15
manually merged wip-more-testing

lobingera:
OK. This initially started to have more testing (coverage) only but now:

- more sample to have more testing coverage and some example of use
- tests organized in test sets
- cleared 0.6dev deprecation warnings
- ~89% test coverage (up from 55)


lobingera:
@JuliaGraphics/collaborators Anyone willing to pull? Or to comment on this?

timholy:
I think this is great work. On a quick scan, the only thing I found to quibble about was indentation and at least one file with a lot of blank lines at the end. My personal view is that such things are not a big deal, but perhaps I'll leave this open in case anyone else objects. If no one gets back to you soon, you can merge, right?

test/test_stream.jl
```diff
@@ -13,4 +13,13 @@ pipe = Base64EncodePipe(buf)
 write_to_png(c,pipe)
 close(pipe)
 # Catch short writes
-@assert length(takebuf_array(buf)) > 200
+
+    if VERSION >= v"0.6.0-dev.1954"
+        str = String(take!(buf))
```
tkelman: can use Compat 0.9.5 for this?

lobingera: I agree. I wasn't looking into Compat on this.

lobingera:
@timholy You're right with the whitespace, i'll take a look.

test/test_stream.jl
```diff
@@ -12,14 +12,10 @@ buf = IOBuffer()
...
 # Catch short writes
 
-    if VERSION >= v"0.6.0-dev.1954"
-        str = String(take!(buf))
-        str_data = Vector{UInt8}(str)
-    else
-        str = takebuf_string(buf)    
-        str_data = str.data
-    end        
+@compat str = String(take!(buf))
```
tkelman: you don't actually need the `@compat` macro for this since it isn't a syntax rewrite, these functions are just exported by Compat

lobingera: Put it like this: I have some trouble to understand the Compat manual ...

tkelman: It's not entirely unique to Compat, it's whether or not something is valid syntax on older versions of Julia. If it is, there's no need for a macro.

lobingera:
@JuliaGraphics/collaborators To sum it up: I consider this ready from my side. But i understand nobody wants to merge before 0.6 testing is passing.

samples/sample_alpha_paint.jl
```diff
@@ -7,7 +7,7 @@ cr = CairoContext(c);
 
 save(cr);
 
-z = Array(RGB24,2,2)
+z = Array{RGB24},2,2)
```
tkelman: missing open paren

REQUIRE
```diff
@@ -1,5 +1,5 @@
 julia 0.3
-Compat 0.8.0
+Compat 0.10
```
tkelman: don't think you're using `.=` or `transcode` ?

lobingera: Sorry, i can't follow: Is 0.10 too aggressive? What is your recommendation?

tkelman: if you aren't using any features that weren't present in 0.8, then the bound should be left at 0.10

it depends if you're using any newer features, can you remember why you increased this?

lobingera: I cannot remember clearly, but i think something with takebuf_string and 0.6

tkelman: ah if you used take!, that was 0.9.5

PR #169, opened 2017-02-14
add unexported vars storing pango/cairo version literals

bjarthur:
i would add similar for gobject and pangocairo but there doesn't seem to be a way to query their version at runtime.

### DIFF
--- a/src/Cairo.jl
+++ b/src/Cairo.jl
@@ -13,6 +13,19 @@ using Colors
 importall Graphics
 import Base: copy
 
+libcairo_version = VersionNumber(unsafe_string(
+      ccall((:cairo_version_string,Cairo._jl_libcairo),Cstring,()) ))
+libpango_version = VersionNumber(unsafe_string(
+      ccall((:pango_version_string,Cairo._jl_libpango),Cstring,()) ))
+if !is_windows()
+    libpangocairo_version = VersionNumber(unsafe_string(
+          ccall((:pango_version_string,Cairo._jl_libpangocairo),Cstring,()) ))
+    libgobject_version = VersionNumber( 
+          unsafe_load(cglobal((:glib_major_version, Cairo._jl_libgobject), Cuint)),
+          unsafe_load(cglobal((:glib_minor_version, Cairo._jl_libgobject), Cuint)),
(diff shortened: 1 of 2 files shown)


lobingera:
a) Glib has https://developer.gnome.org/glib/2.50/glib-Version-Information.html#glib-check-version and Gobject follows, as it's a part of Glib
b) the same applies for pangocairo which is an integral part of pango

bjarthur:
thanks.  but calling `glib_major_version` is crashing the windows appveyor builds.  works fine on linux and mac.  should i add `if !is_windows()` logic, or is there a way to get the version on that platform?

lobingera:
> but calling glib_major_version is crashing the windows appveyor builds

afaics in the docs, that's not a function you call, you just access the symbol to get the data, maybe via cglobal?

bjarthur:
i'm using cglobal already to get glib_major_version.  works great on linux and mac.  can anyone with a windows box help debug?

bjarthur:
i added logic to not set `libpangocairo_version` and `libgobject_version` on windows.  `libcairo_version` and `libpango_version` are set on all platforms.

since tests now pass, i will merge in a couple days unless someone volunteers to debug the windows inconsistencies.

lobingera:
@bjarthur What are you actually doing with these version numbers?

src/Cairo.jl
```diff
@@ -13,6 +13,19 @@ using Colors
 importall Graphics
 import Base: copy
 
+libcairo_version = VersionNumber(unsafe_string(
+      ccall((:cairo_version_string,Cairo._jl_libcairo),Cstring,()) ))
+libpango_version = VersionNumber(unsafe_string(
+      ccall((:pango_version_string,Cairo._jl_libpango),Cstring,()) ))
+if !is_windows()
+  libpangocairo_version = VersionNumber(unsafe_string(
```
tkelman: most of the rest of the package uses 4 space indent

bjarthur:
they're useful for debugging.  i recently had cairo work on one machine and not another so coded up how to query the version info to see if that was the cause.  figured i'd share with everyone else to make it easier in future.

note also the use of version info in `test/test_speed.jl`.

plus other packages, `Gtk.jl` for example, have similar variables.  i think it's generally useful.

PR #175, opened 2017-03-05
introducing mesh-grid pattern functions

lobingera:
manually merged #83, updated to v0.5 syntax, added meshgrid example


src/Cairo.jl
```diff
@@ -30,14 +30,23 @@ end
 
 include("constants.jl")
 
+# check library version, error message if too low
+cairo_version = ccall((:cairo_version,Cairo._jl_libcairo),Int32,())
+
+if cairo_version < 11200
+    error("cairo version " * string(cairo_version) * ", less then 1.12.00")
```
tkelman: this should be done as a bindeps validation check, or make the extra features disabled if the library is too old

lobingera: first part: I tend to agree, however i don't see how BinDeps could check cairo_version itself?
part two: ... i remember a lengthy discussion about conditional exports ...

It's not nice, but the best i could come up with.

tkelman: always erroring at package import time doesn't seem necessary if only the new functionality requires recent versions?

tkelman: see how SCS.jl or Cbc.jl manage bindeps version validation checks in their build.jl scripts. though here it might just be better to skip the new functions if too old (perhaps with a warning?)

lobingera:
@cormullion and @timholy, much more simple: travis linux doesn't have a 1.12 ...

lobingera:
@cormullion; there's hope, Travis-ci is supporting 14.04 in beta, i just need to learn how to configure. For 9c3f282 the appveyor (windows build) is already succeeding in 0.4/0.5, on travis the OSX passed yesterday.

cormullion:
Cool! (I won't ask why Travis is so behind the times.) We could have a new release of Cairo.jl soon, which would be awesome. 😀

lobingera:
@cormullion, for the release i'd also try to include https://github.com/JuliaGraphics/Cairo.jl/pull/167 which then should us bring to 0.6

tkelman:
Given people still do occasionally use ubuntu 12.04 or centos 6 and other old distros, I still think it would be better to have the package partially functional rather than not install at all there. setting travis to 1404 would still be valuable for testing the newer pieces.

lobingera:
@tkelman afair exports cannot be done conditionally. So the function - encapsulating the ccall - itself needs to be available in two versions (as we want to avoid a version test inside the function), one real call, one 'dummy' and assigned at build time. Via macro?
Do you know an example of that behaviour?

tknopp:
@lobingera: don't you just have to remove the build check and everything will be fine? The tests could then explicitly ask for the lib version number and only run the tests if the version is high enough

lobingera:
@tknopp, could be an good idea. Still, if we have 1.12 (and 1.14) features, we need travis testing.

tknopp:
step 1:
have a variable within Cairo.jl that stores the version info:

see: https://github.com/JuliaGraphics/Gtk.jl/blob/master/src/Gtk.jl#L46

Step 2: Only run test if the feature is there:

https://github.com/JuliaGraphics/Gtk.jl/blob/master/test/gui.jl#L739

Then travis should be happy on any Cairo version.

lobingera:
step1 is already there.

step2 is managable. At the moment
https://github.com/JuliaGraphics/Cairo.jl/blob/master/test/runtests.jl#L43-L48
runs everything that is in samples, starts with sample_ and ends with .jl.

tknopp:
Looks good!

lobingera:
@tknopp, i'll try to make some sense of the travis.yml later today. I'm planning with running 0.4/0.5/nightly on linux on trusty and 0.5 on precision additionally (so we'll have a <1.12 testcase). OSX will stay like this.

deps/build.jl
```diff
@@ -4,6 +4,14 @@ using BinDeps
 
 using Compat
 
+# check for cairo version 
+function validate_cairo_version(name,handle)
+    f = Libdl.dlsym_e(handle, "cairo_version")
+    f == C_NULL && return false
+    v = ccall(f, Int32,())        
+    return v > 11000
```
tkelman: which functions require 1.10 ?

lobingera: Good guestion -> https://www.cairographics.org/manual/index-1.10.html but we don't much of them... 

I have to admit, i just changed the version number to something like all-pass as cairo's earlier than 1.10 are really hard to get. Would you insist on dropping this completely?

tkelman: This would be a new restriction, right? I'll have to spin up a docker container and test for sure, but for example it looks like CentOS 6 is at Cairo 1.8.8, judging by https://rpmfind.net/linux/rpm2html/search.php?query=cairo.

lobingera: I see, i'll drop this completely before the release.

PR #177, opened 2017-03-12
Fix a couple depwarns, bump REQUIRE to 0.4

ararslan:
The package is no longer tested on 0.3 and a new version will not be able to be registered without dropping 0.3 support anyway, so this bumps the REQUIRE to 0.4. I've also fixed the `typealias` depwarns on 0.6.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
-julia 0.3
+julia 0.4
 Compat 0.10
 Colors
 BinDeps 0.3.21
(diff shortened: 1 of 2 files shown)


lobingera:
@ararslan 
> new version will not be able to be registered without dropping 0.3 support anyway

Why?

ararslan:
@lobingera It's a rule of METADATA; the minimum Julia version allowed for packages now is 0.4.0.

PR #181, opened 2017-03-20
Quote shell wildcards in the build process

ararslan:
Avoids deprecation warnings on Julia 0.6.

I tried to do the inner constructor syntax too but I only have access to the GitHub web UI at the moment and it kept crashing whenever I tried to make that change. 😕

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -149,9 +149,9 @@ provides(BuildProcess,
             FileRule(joinpath(prefix,"lib","libpng15.dll"),@build_steps begin
                 `cmake -DCMAKE_INSTALL_PREFIX="$prefix" -G"MSYS Makefiles" $pngsrcdir`
                 `make`
-                `cp libpng*.dll $prefix/lib`
-                `cp libpng*.a $prefix/lib`
-                `cp libpng*.pc $prefix/lib/pkgconfig`
+                `cp 'libpng*.dll' $prefix/lib`
+                `cp 'libpng*.a' $prefix/lib`
+                `cp 'libpng*.pc' $prefix/lib/pkgconfig`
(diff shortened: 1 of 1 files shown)


tkelman:
Putting quotes around a wildcard means searching for files that have literal `*` in their names, I don't think this is right

ararslan:
That's kind of what I thought but that's how someone recommended that I do it, so I ran with it. I don't know how else to get the wildcard behavior while also avoiding the depwarn...

tkelman:
would a backslash on the star do the job? that deprecation is pretty broken and annoying.

tkelman:
does this behave as a wildcard though?

ararslan:
> would a backslash on the star do the job?

Doesn't seem to do it in the REPL.

> that deprecation is pretty broken and annoying.

Yeah...

> does this behave as a wildcard though?

You know, I'm actually not sure. With the quoting it actually seems not to, but oddly enough the build succeeds anyway.

tkelman:
This code path probably isn't tested.

ararslan:
Perhaps it isn't needed?

ararslan:
So today I learned that shell expansion has _never_ been performed for `*` in `Command` objects in _any_ Julia version. So unless BinDeps does something weird with `Command` literals, these commands that hit the depwarn actually _cannot_ be necessary.

ararslan:
To quote a guy much smarter than me, "If `cp` is interpreting the `*` specially itself (like the `find` command), then it’ll keep working. If it isn’t, then it was always broken."

So I guess this PR should be fine?

tkelman:
these blocks are for building from source on windows, which is rarely if ever going to work on user's systems. we should probably delete these.

PR #186, opened 2017-05-01
WIP recording and recording to script

lobingera:


test/runtests.jl
```diff
@@ -183,34 +184,61 @@ end
...
 
-    # if Cairo.libcairo_version >= v"1.12.0"
-    #     output_file_name = "a.cs"
-    #     surf = CairoScriptSurface(output_file_name,512,512)
-    #     hdraw(surf,64,8,4) 
-    #     finish(surf)
+    if Cairo.libcairo_version >= v"1.12.0"
 
-    #     @test isfile(output_file_name)
+        if ~is_windows()
```
andreasnoack: Leaving some comments here would probably have saved me a couple of hours. Are you sure this has anything to do with Windows? I'm getting test errors here on Mac. It seems that Cairo only writes to the file when the finalizer [registered here](https://github.com/JuliaGraphics/Cairo.jl/blob/master/src/Cairo.jl#L435) is run. Kind of weird design on their end but I guess we'd have to change on our end to get something reliable. I.e. it seems to me that we can't use finalizers here but would need to explicitly call `destroy` in e.g. https://github.com/JuliaGraphics/Cairo.jl/blob/master/src/Cairo.jl#L435 instead of relying on `s` being garbage collected.

lobingera: What are you trying to do?
Afair there was something windows specific about libuv and opening and closing on files, so writing to a file and trying to see if it exists (in my timezone it's close to midnight, so my mind isn't fully operational) fails. btw: cairo documentation anyway recommends (but i read this way later than implementing here) not to use the file interface, but the stream.

andreasnoack: I'm trying to get the tests to pass on master and just managed to do so by adding `gc()` calls after `destroy`.

lobingera: I see (i'm working on the same, but not all the time). Could you please open an issue, we can discuss there?

PR #188, opened 2017-05-10
Use Libdl.dlext instead of old BinDeps method

musm:


lobingera:
looks good to me, how was this recognized as problem?

musm:
It's about to get deprecated. It's a 3 year old method in BinDeps
https://github.com/JuliaLang/BinDeps.jl/pull/284

PR #190, opened 2017-05-24
trying to increase test coverage by excluding code that not applies to OS

lobingera:


appveyor.yml
```diff
@@ -4,8 +4,8 @@ environment:
   - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x64/0.4/julia-0.4-latest-win64.exe"
   - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x86/0.5/julia-0.5-latest-win32.exe"
   - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x64/0.5/julia-0.5-latest-win64.exe"
-  - JULIA_URL: "https://julialangnightlies-s3.julialang.org/bin/winnt/x86/julia-latest-win32.exe"
```
ararslan: Why remove testing on nightly?

src/Cairo.jl
```diff
@@ -98,9 +98,11 @@ export
     # push+pop group
     push_group, pop_group
 
-function write_to_ios_callback(s::Ptr{Void}, buf::Ptr{UInt8}, len::UInt32)
-    n = ccall(:ios_write, UInt, (Ptr{Void}, Ptr{Void}, UInt), s, buf, len)
-    @compat Int32((n == len) ? 0 : 11)
+@static if is_apple()
```
ararslan: `@static` isn't needed at the top level. Also the `@compat` in this function is unnecessary.

appveyor.yml
lobingera: it's rather like: adding testing on recent 0.6. the wip (work in progress) applies

src/Cairo.jl
lobingera: actually i wanted to test if `@static` can be used to exclude code from the coverage calculation. I think the compat still applies for <0.5

PR #191, opened 2017-06-01
Add CairoPSSurface

davidanthoff:


lobingera:
1) what are you missing from the EPS?
2) and if PS is really needed, please add the DSC interface

davidanthoff:
> what are you missing from the EPS?

I'm adding various file export functions to [VegaLite.jl](https://github.com/fredo-dedup/VegaLite.jl) and it would be nice if we could support both ``*.eps`` and ``*.ps`` files. It is not a huge need, but it didn't seem complicated to add.

> and if PS is really needed, please add the DSC interface

I don't understand. We don't need that for things to work, right? It would of course be nice if someone wants to add that, but we don't really need this if all we want to do is add PS support, right?

lobingera:
EPS and PS are using the same rendering model and technically the same language PostScript. The main difference is how the content is 'packaged' into a file. EPS is rather a single page that can be rendered independently and PS is rather a multi-page document with e.g. DSC comments and information for physical printers (like paper selection and color model). 
So usually when people ask for PS support they talk about supporting all this.

I'm fine with merging this and putting DSC into another PR. Still, the 32bit windows test failures should be clarified.

davidanthoff:
I think the 32bit windows test failures are unrelated to this PR, ``master`` seems to have these failures as well.

lobingera:
I agree, but i cannot do a release with this failing.

davidanthoff:
I won't be able to sort out those build failures. So maybe best strategy is to merge this here, and then hope that someone comes along and fixes the build problems? I don't need the PS support urgently, so if that takes a while, no biggy.

lobingera:
@davidanthoff Offtopic: Where in VegaLight does Cairo show up?

davidanthoff:
fredo-dedup/VegaLite.jl#17, not yet merged and I don't know whether @fredo-dedup is on board. Generally speaking, vega-lite can produce SVG or PNG files. The PNG file stuff only works if you have Cairo installed, but cairo for nodejs, not julia. That seems really complicated, so I'm only using the native svg generation part of vega-lite in that PR.

I then use Cairo.jl to convert that svg into all sorts of other formats, both to implement that various ``show`` methods for all the mime types, and to implement a ``savefig`` function.

PR #193, opened 2017-06-03
take ios stream callback out

lobingera:
it looks like, this was still in, because the standard stream was not working. Now with successful testing of all OS variants it looks like not needed.


tkelman:
There's still one more usage of this in CairoPSSurface

lobingera:
that was effect of a race condition, the PR for including PS surface was not aware that this will be taken out. Do you object to keep this in the current release?

tkelman:
Is it going to break something? It didn't cause a test failure, but do you have coverage of the problematic line?

lobingera:
The whole point in taking the ios streams out was about: it's not tested because setting up an ios stream seems to be something special only in iOS and even on iOS the regular (uv) stream work.

lobingera:
... and i don't think it will break ...

lobingera:
forget my previous comment, i'll update PSSurface on master.

tkelman:
I don't think ios in that naming is referring to an operating system at all, it's an internal name for a part of the io system that is sort of related to / interacts with libuv but is also a bit separate - I think it's part of the femtolisp parser.

lobingera:
@tkelman I cannot really follow. The callback was named write_to_ios_callback was ccall-ing a :ios_write - without a library reference. And from the history of commits it looked to me like: this was an OS specific implementation and on 'other' OS it was done via uv. Now uv works on all platforms and is tested.

Do you think we should bring back the stream::IOStream in addition to {T<:IO}(stream::T ?

tkelman:
It may have been that it was only needed on some OS's, but that's not what the "ios" in the name stands for as far as I know. If you checked blame a bit to find who introduced it, they'd be the best person to ask.

PR #196, opened 2017-08-05
WIP: test gcc dll replacement on appveyor

tkelman:
not sure if this will work, but this may be a way to deal with the win32 issue

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -39,7 +39,7 @@ build_script:
   - IF EXIST .git\shallow (git fetch --unshallow)
   - C:\projects\julia\bin\julia --startup-file=yes -e "versioninfo();
       Pkg.add(\"WinRPM\"); Pkg.checkout(\"WinRPM\", \"tk/gccdll\");
-      Pkg.clone(pwd(), \"Cairo\"); Pkg.build(\"Cairo\")"
+      Pkg.clone(pwd(), \"Cairo\")"
 
 test_script:
-  - C:\projects\julia\bin\julia -e "Pkg.test(\"Cairo\")"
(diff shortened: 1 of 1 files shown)


lobingera:
@tkelman Do you think this can be included in mainstream WinRPM soon?

tkelman:
it's a bit risky since the opensuse repo can change at any time, if anything breaks it may result in julia not starting correctly. maybe we could do it only on 32 bit?

lobingera:
@tkelman 
> maybe we could do it only on 32 bit?

yes. 

Can you explain what error/failure you're working around? In my own quest to see why the 32bit fails it looked like the library is installed, but never recognized by julia

tkelman:
The short version is the opensuse build service upgraded to gcc 7 a few months ago, but the Windows Julia binaries are still built (cross-compiled from cygwin on the buildbots) using gcc 5. Evidently there are a few new functions in libgcc in gcc 7 that get used by cairo, but the libgcc that Julia includes does not have.

lobingera:
@tkelman Would you recommend to merge this, now?

tkelman:
Not really, not in this exact form. Could maybe start doing this over in WinRPM's build script but only on 32 bit (since it introduces a risk of upstream changes from opensuse breaking Julia installs unpredictably).

I'm no longer actively doing Julia development, so addressing this issue should probably be undertaken by someone else who regularly uses and tests things on Windows. I don't think there's anyone who fits that description at the moment, at least in terms of active people with commit access to base Julia. Maybe BinDeps2 will have a fully working Windows cross-compilation story eventually.

lobingera:
BinDep2 development already started?

ararslan:
> BinDeps2 development already started?

https://github.com/JuliaPackaging/BinDeps2.jl

giordano:
This can probably be closed now, I guess

PR #197, opened 2017-08-06
WIP: preparation for v0.7

lobingera:


test/runtests.jl
```diff
@@ -210,7 +211,7 @@ end
 
     if Cairo.libcairo_version >= v"1.12.0"
 
-        if ~is_windows()
+        if ~Sys.iswindows()
```
tkelman: will need to raise Compat min version, in either test/REQUIRE or REQUIRE

ararslan: It will also have to be invoked as `Compat.Sys.iswindows()`

lobingera: @ararslan But that might be a Compat documentation issue 

> is_apple, is_bsd, is_linux, is_unix, and is_windows are now Sys.isapple, Sys.isbsd, Sys.islinux, Sys.isunix, and Sys.iswindows, respectively. These are available in the Compat.Sys submodule.

reads for me like using Compat.Sys and apply Sys.

ararslan: Yeah, that's a failure of the documentation.
```
julia> using Compat.Sys

julia> Sys.iswindows()
WARNING: both Sys and Base export "Sys"; uses of it in module Main must be qualified
ERROR: UndefVarError: Sys not defined
```
Though actually it appears to work (at least on 0.6) with `import Compat.Sys`.

lobingera:
NOTE: this will be superceeded by #216

Issue #202, opened 2017-11-02
Possible bug in  mesh_pattern_set_corner_color_rgba()?


cormullion:
Is it possible that the `_rgba` version of the `mesh_pattern_set_corner_color_` function has a copy/paste error, calling `:cairo_mesh_pattern_set_corner_color_rgb` rather than an alpha-friendly `:cairo_mesh_pattern_set_corner_color_rgba`?

Line 977 - 984:

```
function mesh_pattern_set_corner_color_rgb(pat::CairoPattern, corner_num::Real, red::Real, green::Real, blue::Real)
    ccall((:cairo_mesh_pattern_set_corner_color_rgb, _jl_libcairo),
                    Void, (Ptr{Void},Int32,Float64,Float64,Float64),pat.ptr,corner_num,red,green,blue)
end

function mesh_pattern_set_corner_color_rgba(pat::CairoPattern, corner_num::Real, red::Real, green::Real, blue::Real, alpha::Real)
    ccall((:cairo_mesh_pattern_set_corner_color_rgb, _jl_libcairo),
                    Void, (Ptr{Void},Int32,Float64,Float64,Float64,Float64),pat.ptr,corner_num,red,green,blue,alpha)
end
```

lobingera:
yes. Will you do a PR?

PR #204, opened 2017-11-04
Change second cairo_mesh_pattern_set_corner_color_rgb to ..._rgba

cormullion:
To fix https://github.com/JuliaGraphics/Cairo.jl/issues/202


lobingera:
@cormullion, i'll accept the PR. The win32 situation is a known problem #196, the MacOS seems to be major problems with Homebrew.

PR #220, opened 2018-02-22
Fix and add test for reset_transform

kdw503:
A recent Cairo update seems to have caused troubles for ImageView due to `reset_transform` from `Graphics` not being imported. It's fixed by the `importall` on master; this just adds a test to make sure it doesn't regress.

### DIFF
--- a/src/Cairo.jl
+++ b/src/Cairo.jl
@@ -13,7 +13,7 @@ using Colors
 
 import Graphics
 using Graphics: BoundingBox, GraphicsContext, GraphicsDevice
-import Graphics: arc, clip, close_path, creategc, fill_preserve, height, line_to, move_to, new_path, new_sub_path, paint, rectangle, rel_line_to, reset_clip, restore, rotate, save, scale, set_dash, set_line_width, set_source, set_source_rgb, set_source_rgba, stroke, stroke_preserve, textwidth, translate, width, circle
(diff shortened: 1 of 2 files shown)


kdw503:
Test error fixed (I was using an old version of master, it turns out there was still a bug to fix).

lobingera:
@kdw503 

> It's fixed by the importall on master;

lobingera:
@kdw503 Sorry for the confusion. Could you take a look at the actual test (see it fail in above)? It looks like you're testing Graphics.reset_transformation and Graphics is not even imported.

kdw503:
Sorry, forgot to use Graphics. I added 'using Graphics'. But still has reset_transform error.

timholy:
This is a real bug (ImageView is no longer working) and a real fix. Graphics specifies that implementations of `reset_transform` are [required](https://github.com/JuliaGraphics/Graphics.jl/blob/3183be960f53f7782c7ae6134ea25bb5905d7f77/src/Graphics.jl#L391). Cairo defines an internal `reset_transform` (and many other `Graphics` stubs), but unlike the other stubs Cairo doesn't `import` the function from `Graphics`. If you've said `using Graphics` then boom everything is broken.

timholy:
Ref https://github.com/JuliaImages/ImageView.jl/issues/139

timholy:
Thanks!

timholy:
Possible to tag a new release?

lobingera:
I'm having some connectivity problems right now, but I think you have the rights also, just do a 0.5.1

kdw503:
Thank you all.

Issue #223, opened 2018-03-12
Labels not showing up in ProfileView


antoine-levitt:
See https://github.com/timholy/ProfileView.jl/issues/91. I'm opening an issue here because pinning Cairo to 0.4 fixes the issue.

lobingera:
I could reproduce this, locally with linux, julia 062. As GtkReactive shows no fail messages or errors on importing, it might be hard to debug.

timholy:
Fixed by #254

PR #227, opened 2018-05-29
remove Homebrew check

jonathanBieler:
This doesn't work on v0.7 because `Pkg.installed` doesn't take a String anymore. While it could be fixed with Compat I propose to remove the check since the error message for `using Homebrew` is very similar and we don't do the check for windows. Homebrew is also in `REQUIRE` so it should be installed.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -3,7 +3,6 @@ using BinDeps
 using Compat
 import Compat.Libdl
 import Compat.Sys
-import Compat.Pkg
 
 @BinDeps.setup
 
@@ -40,9 +39,6 @@ if Sys.iswindows()
 end
 
 if is_apple()
-    if Pkg.installed("Homebrew") === nothing
-        error("Homebrew package not installed, please run Pkg.add(\"Homebrew\")")
-    end
     using Homebrew
     provides( Homebrew.HB, "cairo", cairo, os = :Darwin )
     provides( Homebrew.HB, "pango", [pango, pangocairo], os = :Darwin, onload =


lobingera:
@jonathanBieler Looks reasonable. Do you think the failing test (travis, nightly, macos) is related?

tknopp:
@lobingera I can confirm that this make Cairo usable under 0.7 / Mac again. travis seems to be another issue

aviks:
Is this worth a new release please?

lobingera:
Is this urgent and solving someone's problem? I'd like to update some deprecation warnings also, get 0.7 testing badge and find a solution for https://github.com/JuliaGraphics/Cairo.jl/issues/223 which seems to be a side effect of the replacement of 'importall'. Time horizon: This weekend.

aviks:
Yes,  of course. I was stuck, but a `Cairo#master` has let me progress for the moment. Thanks!

aviks:
And, to add, I've been troubled by #223 as well (on 0.6, mac), so a fix for that would be much appreciated!

tknopp:
Transition to 0.7 seems to require a step-by step approach. First the "leaf-packages" need to migrate 0.7 and then make a release. As a depending package (like Winston.jl) first the dependencies need to get ready. Cairo is of course pretty close to a leaf package.

aviks:
> Cairo is of course pretty close to a leaf package

And hence I'm here. @tknopp I'll ask about Gtk next. :)

tknopp:
@aviks: I kind of feared that. Fortunately @jonathanBieler has (almost?) finished the port: https://github.com/JuliaGraphics/Gtk.jl/tree/teh/partial0.7

lobingera:
@tknopp ,@aviks, 0.5.2 is tagged. I had no time to look into #223.

PR #233, opened 2018-08-06
Ci config for 0.7

davidanthoff:


davidanthoff:
I think this still needs some tweaks to work properly. The first is that for the new official appveyor scripts to work, the appveyor project name (in the appveyor settings on their website) needs to be ``Cairo.jl``, otherwise the script doesn't work. @lobingera is that something you can change? I believe that explains all the appveyor failures right now.

PR #234, opened 2018-08-06
Some 0.7 compat

davidanthoff:


davidanthoff:
Things still don't work on 0.7 with this. The remaining offending line is [here](https://github.com/JuliaGraphics/Cairo.jl/blob/2ec9f9d7a80022fb1403ff600588a82b611178d4/src/Cairo.jl#L302). That line references a function called ``write_to_ios_callback`` that is nowhere defined. On 0.7 the package won't load with this. @lobingera I believe you removed that function at some point, but I guess there is a need to somehow update ``CairoPSSurface`` to work without it. Probably easiest if you figure that out?

lobingera:
I got an answer on my cfunction question, I'll visit this anyway. I thought I had dropped iOS streams anyway, because the standard stream work everywhere and nobody uses iOS special

lobingera:
Note to myself: read the code part of the pr also - even when checking from mobile ...

lobingera:
This was included in https://github.com/JuliaGraphics/Cairo.jl/commit/14e0711465f73c483de76351483bee3a874eba91

PR #244, opened 2018-08-15
Update build.jl

tknopp:


deps/build.jl
```diff
@@ -92,6 +91,7 @@ provides(Zypper,
 
 const png_version = "1.5.14"
 
+#=
```
tkelman: commenting this out breaks the sources provider. hold off on immediately merging and releasing your own PRs?

tknopp: sorry @tkelman: I just wanted to help out so that people can use Julia 1.0. Please revert if the from source builds are important.

tkelman: they're necessary on any distro whose package manager is not listed, or when people don't have sudo. the world is not ubuntu. they'll be needed in the same cases on 1.0 too, there's nothing version of julia specific about this

tknopp: It was my fault and I apologize! Would be great if the issue can be fixed for 1.0, but in the mean time https://github.com/JuliaGraphics/Cairo.jl/pull/245 could be a step in between. Please have a look.

Issue #230, opened 2018-07-21
Cairo won't install on clean v0.6 system (macOS High Sierra, Julia v0.6.4)


cormullion:
I tried to install Cairo, on a new installation. It fails with a pangocairo error.

```
===============================================================[ ERROR: Cairo ]===============================================================

LoadError: Provider Homebrew.HB failed to satisfy dependency pangocairo
while loading /Users/me/.julia/v0.6/Cairo/deps/build.jl, in expression starting on line 161

==============================================================================================================================================

===============================================================[ BUILD ERRORS ]===============================================================

WARNING: Cairo had build errors.

 - packages with build errors remain installed in /Users/me/.julia/v0.6
 - build the package(s) and all dependencies with `Pkg.build("Cairo")`
 - build a single package by running its `deps/build.jl` script

==============================================================================================================================================
INFO: Package database updated

julia-0.6> 

```

Some pangocairo libraries were installed:

```
 ~/.julia/v0.6/Homebrew/deps/usr/lib  $ ls *pan*
libpango-1.0.0.dylib		libpangocairo-1.0.0.dylib	libpangoft2-1.0.0.dylib
libpango-1.0.a			libpangocairo-1.0.a		libpangoft2-1.0.a
libpango-1.0.dylib		libpangocairo-1.0.dylib		libpangoft2-1.0.dylib
```

lobingera:
Might be a BinDeps (in 0.6.4) or a Homebrew problem; i've seen similar - not finding the library - also on linux. This is really hard to reproduce and debug. My (local) workaround is: take care that libraries are installed and write deps.jl manually.

cormullion:
So it might be possible to borrow a `deps.jl` file from somewhere and put it into `Cairo/deps/`, and point it to the libraries it needs? OK, I'll have a go at that this week.

CiaranOMara:
I have the same issue.

matthieugomez:
Same issue.

lobingera:
i see this reproduced in https://travis-ci.org/JuliaGraphics/Cairo.jl/jobs/401860559

cormullion:
I'm a bit puzzled why this started recently, I thought Julia v0.6 was stable, and v0.7 was where all the instability was. Cairo seems to install happily on version v0.7, and I'm even able to use the v0.7 Cairo bindeps.jl with v0.6.

fipelle:
Same problem on a clean v0.6.4.

[...]

PR #250, opened 2018-08-22
Fix Homebrew issues on 0.7 and 1.0

tknopp:
thanks to @affans
https://github.com/JuliaGraphics/Cairo.jl/issues/230#issuecomment-414886635

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,8 +34,6 @@ matrix:
        - language: julia
          julia: 1.0
          os: osx
-    allow_failures:
-       - julia: 1.0
 notifications:
     email: false
 after_success:
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -39,6 +39,7 @@ end
 
 if Sys.isapple()
     using Homebrew
+    Homebrew.add("graphite2")
     provides( Homebrew.HB, "cairo", cairo, os = :Darwin )
     provides( Homebrew.HB, "pango", [pango, pangocairo], os = :Darwin, onload =
     """


tknopp:
This makes me so happy. Appveyor failure is some download error. That passed before I modified travis.

affans:
Glad i could help. I needed all of this for `vegalite` plotting.

tkelman:
This should be fixed at the homebrew recipe level. This is like fixing a missing dependency by manually calling Pkg.add instead of adding it to REQUIRE where it belongs

tknopp:
Yes that is clear. It just needs someone with homebrew knowledge (i.e. not me) to do this. @staticfloat maybe.

lobingera:
@tkelman That's why it's called 'Fix', not 'Solution'...

tknopp:
This discussion leads to nowhere. I find it pretty unsatisfying that the people putting energy into this project are blamed for their work. We don't need to justify ourself to someone who does not use Julia anymore. 

@timholy: I would really like to get support from you in this matter. We got travis all green with this patch and terms like `messy hack workaround` are in my opinion not appropriate here (although I have to admit not being a native speaker).

tkelman:
The problem is more or less here. https://github.com/Homebrew/homebrew-core/blob/83dda4fe2614ad89c3d9715ce0a17f3f580874bf/Formula/harfbuzz.rb#L25

graphite2 is listed as a "recommended" dependency of harfbuzz by homebrew, but the binaries of harfbuzz they're distributing are built linked to graphite2. If you download homebrew's harfbuzz bottles directly and look at the dylibs with otool you should be able to verify this. Either homebrew should be building bottles without linking against recommended dependencies, or the dependencies that are linked against should be handled as full dependencies when installing bottles. If the brew cli already implements this logic for recommended dependencies, then the bug may be in Homebrew.jl's treatment of them.

lobingera:
>  If you download homebrew's harfbuzz bottles directly and look at the dylibs with otool you should be able to verify this.

Can i do this on Linux?

tkelman:
Possibly with ObjectFile.jl? Check what the mac build image for BinaryBuilder uses for otool-like functionality when inspecting linkages. Otherwise it's usually violating the mac sdk's eula to use it on non-apple hardware, annoyingly. http://www.newosxbook.com/tools/jtool.html might be a workable alternative

lobingera:
I see. Still the responsibility to do this correctly is with the homebrew-core formular people? btw: Why are other brew users are not seeing these problems?

tkelman:
I'm not sure what brew's behavior would be in a totally clean environment if you do `brew install harfbuzz` - if it installs graphite2, then Homebrew.jl and the brew cli apparently differ in behavior on recommended dependencies.

lobingera:
@tkelman I borrowed a Macbook and otools shows me on the libharfbuzz.dylib (download from bottle) a dependency on graphite. The jtool thing isn't compiling on linux

tkelman:
And does brew install harfbuzz in a clean environment pull in graphite2? If so, then Homebrew.jl isn't replicating brew's behavior accurately enough.
________________________________

lobingera:
I couldn't run brew on the other computer. Maybe via travis.

lobingera:
@tkelman Can you explain, what 'wrong' effects on a julia installation could happen if graphite2 is installed manually?

tkelman:
This should not be released as-is, it's an incorrect fix and masking a bigger problem that should be fixed directly - either Homebrew.jl isn't handling recommended dependencies correctly, in which case this isn't the only package that will see incorrect behavior, or the homebrew bottle itself was built wrong. It is a bit unfortunate that it requires debugging on a mac to identify which it is. Hopefully someone who's hitting the issue can assist with that.

The workaround here incorrectly treats graphite2 as if it was manually requested to be installed by the user, rather than only being installed as a dependency of something else.

timholy:
@tkelman, can you just fix it yourself? The key concern is that it appears none of us know how. While I have a very superficial understanding of the issues here, I agree with your sentiments; nevertheless, I assure you we are not going to freeze progress in Cairo.jl indefinitely (see #254) just for a fix that would be "more ideal."

At a minimum, since Cairo isn't the fundamental problem it would be good to report our concerns in the appropriate location. I'm not the right person to do that.

tkelman:
Without a mac, no I can't determine the correct fix either. I'm fine releasing 254 without this. Some of the individuals who are hitting the issue that this works around should be able to assist in identifying the actual problem. Really anyone with a mac.

timholy:
OK, let's revert this. There are lots of mac users in #230, let's continue the conversation there. @tkelman if you can coach one of them through what they should be trying, that would be a huge help.

tkelman:
The longer this gets left in place, the more users are going to end up in an incorrect state that will interfere with future debugging and installation, given the wrong treatment of the graphite2 dependency here. The "at a minimum" has not been done, and there's shockingly little indication of anyone caring about fixing this the permanent, correct, non-harmful, upstream way. A little bit of iteration (on a mac) through what's happening on 0.6 and 0.7 to identify whether it's Homebrew.jl or the homebrew packaging of cairo/harfbuzz at fault here is all that's necessary. If this incorrect band-aid gets left here indefinitely, is anyone actually going to look into the real problem?

PR #255, opened 2018-09-02
Revert "Fix Homebrew issues on 0.7 and 1.0 (#250)"

timholy:
This reverts commit 2017ef48c1b69af3a81cdadf6292f3853ae597ca. In issue #250 it was noted that the fix this reverts was not the right fix; while in practice it has the benefit of getting things working for people, it's useful to have an audience of mac users held captive (#230) so that someone is motivated to fix this for real. :laughing:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,8 @@ matrix:
        - language: julia
          julia: 1.0
          os: osx
+    allow_failures:
+       - julia: 1.0
 notifications:
     email: false
 after_success:
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -39,7 +39,6 @@ end
 
 if Sys.isapple()
     using Homebrew
-    Homebrew.add("graphite2")
     provides( Homebrew.HB, "cairo", cairo, os = :Darwin )
     provides( Homebrew.HB, "pango", [pango, pangocairo], os = :Darwin, onload =
     """


timholy:
OK if I tag a release?

lobingera:
@timholy I let you decide. If all remaining gtkReactive and https://github.com/JuliaGraphics/Cairo.jl/issues/223 are sucessfully tested; and there is a clear understanding why 0.6 passes on OSX and 0.7/1.0 fail with missing pangocairo.

timholy:
It certainly fixes #223. I can't comment on the OSX issues (I try not to follow *too* many packages or I go insane), which is why I asked you.

lobingera:
Put it this way: I'm not fine in releasing something that's not fully working in testing, if we don't take CI + regression seriously, then how should others learn to do it...

btw1: @davidanthoff just reminded me, that we should bring back win32 testing
btw2: if you release, make it clear in naming and github flag, that it's only an intermediate step and prerelease.

tknopp:
@timholy: This is pretty frustrating. I put a lot of effort to getting Cairo.jl working on Mac (note: I have a Mac) and my motivation was that we provide all users that want Cairo.jl (or Gtk.jl!) an experience that just works. My fix is absolutely harmless. It might shadow the actual problem but it also has no bad side effects. What you are saying in your OP is: Let those Mac users handle this, not our problem. 


@lobingera: I think this is a misinterpretation of what a "release" is. A release is that thing that is visible to people using Pkg.add. You cannot say that thats a prerelease. Its a release. And in addition we have to look at our dependencies as well. For that reason most Julia package have a very fast release packages since releases are cheap. All fixes should be released ASAP (IMHO).

lobingera:
This is exhausting. Give me 1h, i'll put back the fix and release (and make @tkelman unhappy with that).

lobingera:
see #257 https://github.com/JuliaLang/METADATA.jl/pull/17605 and #256

timholy:
I'm fine with that too. The main point is to avoid getting paralyzed with indecision. I'm thrilled if someone decides to fix it in another way.

tkelman:
It does have negative consequences. It's incorrectly solving the problem and incorrectly tracking the dependency. graphite2 should not be treated as if the user manually asked for it to be installed, if it's only needed as a dependency - and if it's only an optional dependency, the fact that it's linked to at all is a bug that needs to be fixed elsewhere.

PR #260, opened 2018-09-03
WIP Use zlib instead of zlib-devel on Windows

davidanthoff:
I have no idea whether that might help with win32, but lets try :)


davidanthoff:
Clearly not a good strategy :)

PR #267, opened 2018-12-22
Add gettext alias for OpenSUSE Zypper

colbec:
OpenSUSE Zypper seems to use both "gettext-runtime" and "intltool" packages to cover gettext. intltool is not installed in OpenSUSE by default. Installing intltool and adding new intltool alias for gettext seems to resolve an issue on my computer where even though gettext-runtime is installed, Julia reports that gettext is not installed. I believe the call to gettext is requiring functionality from the intltool package.


lobingera:
@colbec Could you please check, if you are uptodate with your local branch? The conflict seems to be the at-windows only, which we dropped long time ago

colbec:
Yes I see what you mean. Oddly enough I thought I did a complete removal of Cairo and reinstall into Julia so was confident that I had the right version. I just repeated the process more carefully and this time I seem to have fetched a file with  and not at-iswindows. My first process must have been faulty in some way. But on the other hand I have been quite religiously doing package updates. I am quite puzzled by this. I will withdraw this PR and resubmit.

colbec:
Aha, I just tried to fork the master copy to revise my PR and after forking I get a version with at-windows again. I need to figure out why I am getting an old version. Perhaps because I already have an ancient copy.

PR #280, opened 2019-04-19
Move from REQUIRE to Project.toml

tknopp:
In order to do the next release we need to remove REQUIRE and add Project.toml. It would theoretically possible to still support Julia 0.6 but I strongly ecourage us to not do this.

I already set the version to 0.6.0. Prior to the release, I would update travis and remove all the 0.6 testing.

The release is very important and urgent so that Mac users get a working Cairo / Winston / Gtk on Julia > 0.7. 

@lobingera: Objections?


lobingera:
@tknopp I don't have. Actually i started this morning reading into Pkg about doing this step ...

tknopp:
ok, I will first make CI happy and then prepare the release. I already have done releases to the ne registry. The only that I have not used is TagBot. But lets try, how this works.

davidanthoff:
I would not commit the `Manifest.toml`, I think the general idea now is to not have those in the repo for packages.

davidanthoff:
Also, I should say I would _love_ to see this merged quickly and a new release, the whole "Cairo not working on Mac" situation is a bit of a disaster for us, especially since this has been the case for many weeks by now.

tknopp:
@davidanthoff: Any citeable reference that the Manifest should not be checked in? Last discussions I recall were not clear on that. If so, its a little bit confusing since any Pkg interaction will generate a new one. Maybe we should put it into .gitignore then.

tknopp:
@JuliaRegistrator register()

davidanthoff:
Nothing citeable, and I think you are right that one hears somewhat different things from different folks. But my sense is that most packages don't seem to have it, and if I look at packages from example KristofferC (who after all worked on the package manager) he also doesn't have it.

I think adding it to `.gitignore` seems like a good solution.

jonathanBieler:
I just installed Cairo's master on a clean macOS (which solved the segfault) and it later threw that error:

```julia
(v1.1) pkg> resolve
 Resolving package versions...
ERROR: Unsatisfiable requirements detected for package Cairo [159f3aea]:
 Cairo [159f3aea] log:
 ├─possible versions are: 0.6.0 or uninstalled
 ├─Cairo [159f3aea] is fixed to version 0.6.0
 └─found to have no compatible versions left with Gtk [4c0ca9eb] 
   └─Gtk [4c0ca9eb] log:
     ├─possible versions are: [0.1.0, 0.2.0, 0.3.0-0.3.2, 0.4.0, 0.5.0, 0.6.0-0.6.2, 0.7.0-0.7.10, 0.8.0-0.8.6, 0.9.0-0.9.3, 0.10.0-0.10.4, 0.11.0, 0.12.0, 0.13.0-0.13.1, 0.14.0, 0.16.0-0.16.5] or uninstalled
     └─restricted to versions 0.16.5 by an explicit requirement, leaving only versions 0.16.5
```

I have only Gtk installed (last release). Removing Project.toml solved the problem. I don't understand these dependency issues, maybe Gtk also needs a Project.toml file ?

tknopp:
@jonathanBieler: We certainly need to upgrade Gtk as well, but I wanted to start with Cairo. I have dev-ed Gtk here and it works.

PR #290, opened 2019-08-29
Run tests on Julia 0.7, 1.0, 1.2, 1.3

giordano:
I'm dropping Julia 1.1 as it is not supported any more, having being superseded by Julia 1.2, see https://discourse.julialang.org/t/julia-v1-2-0-has-been-released/27742

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,9 @@ os:
   - osx
 julia:
   - 0.7
-  - 1.1
+  - 1.0
+  - 1.2
+  - 1.3
   - nightly
 notifications:
   email: false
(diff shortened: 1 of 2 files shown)


lobingera:
I understand, but for packages that define infrastructure, some more should be tested.
1.2 is current -> 1.1 + 1.0
I'm wondering if 0.7 should be dropped now, as it's 'just' a variant of 1.0.

giordano:
Ok, so drop 0.7 completely (I'm totally fine with it), and test all versions 1.0 -> 1.3 + nightly?

PR #292, opened 2019-09-02
Use BinaryProvider to install Cairo and Pango

giordano:
This PR installs Cairo and Pango library using [`BinaryProvider.jl`](https://github.com/JuliaPackaging/BinaryProvider.jl).

I think it's good to keep this open for some time to let users play with it.  So, please _**do test this PR**_ and report back any issue you face.  To do this, just run
```
]add https://github.com/giordano/Cairo.jl.git#binary-builder
```
in Julia REPL.  I recommend then to clean the directory `Cairo/deps/` from `build.log` and `deps.jl`, and then you can run `]build Cairo`.

Building these libraries has been a great stress-test for the new `BinaryBuilder.jl` framework that will land in Julia v1.3.  A _huge_ thanks to @staticfloat! :whale2: :santa:   I'll probably update the URLs of some `build.jl` files later this week (I'll use the "official" ones that will be in https://github.com/JuliaBinaryWrappers), but the content of the downloaded tarballs should be the more or less the same.

When merged, this PR fixes #105, fixes #121, fixes #148, fixes #162, fixes #165, fixes #185, fixes #187, fixes #203, fixes #207, fixes #214, fixes #230, fixes #239, fixes #256, fixes #258, fixes #261, fixes #265, fixes #266, fixes #271, fixes #279, fixes #284, fixes #286, fixes #287.  It supersedes #149, #196, #289.

PR #293, opened 2019-09-03
Use Artifacts and JLL packages on Julia 1.3+

staticfloat:
This is an example PR that showcases how, using the capabilities in Julia 1.3+, binary installation can be accomplished both more quickly and with less developer hassle.  No more `build.jl` files, no need for a `Pkg.build()` step at all, in fact.  Simply write `using Glib_jll`, and you will import bindings for `libgobject`, `libglib`, etc...  Similarly, `using Cairo_jll` will import a binding for `libcairo`.  To learn more about these JLL packages, read this [WIP blog post](https://github.com/JuliaLang/www.julialang.org/pull/417/files).  To learn more about the `Pkg.Artifacts` system that this is all built on top of, read the [relevant docs](https://julialang.github.io/Pkg.jl/dev/artifacts/).


staticfloat:
Huzzah!  It's working!

staticfloat:
To help frame this in light of #292, this branch only works on Julia 1.3+, and it uses the new Artifacts installation code which has many improvements over Binary Provider for complicated use cases exactly like Cairo.jl.  Examples of things that using the Artifacts system will get you:

* Upgrading Cairo versions no longer requires a rebuild; because artifacts are stored outside of the package directory, you no longer need to re-install `libcairo` just because a new version of `Cairo.jl` was released.

* Fewer `Pkg.build()` errors; because no code is run upon package install, (and no `deps.jl` file is generated) package install should be much more reliable.

* Smarted dependency handling; Mosè has done a great job handcrafting the `build.jl`, but there's some unfortunate fragility within that file; installing dependencies in the wrong order will break, sometimes you have to disable a safety feature ("isolated dlopen") to get things to unpack properly, etc... JLL packages avoid these problems by doing all necessary inspection at build-time, then encoding the implicit ordering of build products into the dependency graph of JLL packages.  In short, you don't need to worry about this, the complexity is abstracted away from you.

* Better executable product handling; Executables need certain environment variables set to find their dependencies (`LD_LIBRARY_PATH`, `PATH`, etc...) the new JLL packages provide a convenient interface to walk the recursive dependency tree and set up the requisite environment variables without you ever needing to think about it.

For these reasons, I encourage a Julia 1.3+ release of Cairo, however due to the importance of this package a 1.2- branch will undoubtedly be relied upon for a while yet.  Thus, this PR simply makes it easy to launch the 1.3+ branch when the developers are ready.

tknopp:
This is great @staticfloat! 

So this is an alternative to #292 or can #292 first be merged and then after that this one?

Regarding releases, I would prefer that we make simultaneous releases of Cairo and Gtk, once the later has migrated to BB/BP. I am fine making both 1.3 exclusive. But then the release can only be made once 1.3 is out.

staticfloat:
#292 should be merged first, as that is compatible with the largest number of users right now.  IMO, it's best to make a release of Cairo with these new binaries vendored through BinaryProvider, have that be compatible with Julia 1.0-1.2, then split that commit off into a `compat-1.2` branch, and merge this into `master`.  Then, by the time 1.3 is released, we will have Cairo, Gtk, Rsvg, etc... all ready and working on 1.3+.

I do not think we're going to create a 1.0-compatible release for Gtk, we're just going to skip to 1.3+ directly for those packages, because it's a fair amount of developer work to get it working, and we'd rather just make quick headway with the new tools.  `Cairo.jl` was an exception because it's an extremely high-visibility package, and I personally get a lot of mail from people complaining that it's not working on 1.0.  :)

timholy:
I'm fine with going forward with Julia 1.3+.

I suspect Gtk will become the source of your mail now. At least for me, all my reports are of the form "can't use ImageView" or "can't use ProfileView" and that ultimately traces back to Cairo. Now we've moved a big step up the chain. Hoping that the new infrastructure makes Gtk not too bad. Anything we mere mortals can do to help out usefully? This is not an area of strength for me, but I feel bad about begging for help without doing something useful to pitch in.

tknopp:
I agree with Tim, that most Cairo issues reported are likely indirect issues with Gtk. Cairo of course stands on its own, but if you want to draw not only to a file but to the screen, something like Gtk is necessary

lobingera:
@timholy If you don't work on the binaries but want to contribute, there is still something you can do for adapting interfaces on the julia level e.g. https://github.com/GiovineItalia/Gadfly.jl/issues/1309

src/Cairo.jl
```diff
@@ -13,11 +16,11 @@ Base.@deprecate_binding _jl_libpango      Cairo.libpango   false
...
 
 function __init__()
-    check_deps()
-    # On Linux, FreeBSD and macOS we use FontConfig. Set FONTCONFIG_FILE to the
-    # config file we provide.
-    if !Sys.iswindows()
-        ENV["FONTCONFIG_FILE"] = joinpath(dirname(libcairo), "..", "etc", "fonts", "fonts.conf")
+    # On Linux and FreeBSD we use FontConfig. Set FONTCONFIG_PATH only if none
+    # of FONTCONFIG_PATH or FONTCONFIG_FILE is set.
+    if !(Sys.isapple() || Sys.iswindows()) && get(ENV, "FONTCONFIG_PATH", "") == "" &&
```
giordano: Fontconfig is now on all platforms, right?

src/Cairo.jl
```diff
@@ -13,11 +16,11 @@ Base.@deprecate_binding _jl_libpango      Cairo.libpango   false
...
+        get(ENV, "FONTCONFIG_FILE", "") == ""
+        ENV["FONTCONFIG_PATH"] = joinpath(dirname(libcairo), "..", "etc", "fonts")
```
giordano: This should use libfontconfig, as the libraries now live each in their own root

src/Cairo.jl
staticfloat: Mmmm, yes, I believe that is true yes.

src/Cairo.jl
giordano: I had a feeling of déjà vu: https://github.com/staticfloat/Cairo.jl/pull/1

Manifest.toml
```diff
@@ -0,0 +1,256 @@
+# This file is machine-generated - editing it directly is not advised
```
davidanthoff: Remove this whole file?

Project.toml
```diff
@@ -4,14 +4,18 @@ version = "0.8.0"
 
 [deps]
 BinaryProvider = "b99e7846-7c00-51b0-8f62-c81ae34c0232"
```
davidanthoff: We can remove BinaryProvider.jl, right?

Project.toml
```diff
@@ -4,14 +4,18 @@ version = "0.8.0"
...
+Cairo_jll = "83423d85-b0ee-5818-9007-b63ccbeb887a"
 Colors = "5ae59095-9a9b-59fe-a467-6f913c188581"
+Fontconfig_jll = "a3f928ae-7b40-5064-980b-68af3947d34b"
+Glib_jll = "7746bdde-850d-59dc-9ae8-88ece973131d"
 Graphics = "a2bd30eb-e257-5431-a919-1863eab51364"
 Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
+Pango_jll = "36c8627f-9965-5494-a995-c6b170f724f3"
 
 [compat]
 BinaryProvider = "≥ 0.3.0"
```
davidanthoff: Remove `BinaryProvider` here.

Project.toml
```diff
@@ -4,14 +4,18 @@ version = "0.8.0"
...
 Graphics = "≥ 0.1.0"
-julia = "0.7, 1"
+julia = "≥ 1.3.0"
```
davidanthoff: I think this should be `julia = "1.3"`.

Manifest.toml
staticfloat: Do you not want to include a `Manifest.toml`?  It can be helpful to track down why, for instance, improper upper bounds caused a test to fail in the future.

staticfloat:
Alright there was some confusion because I re-named a branch halfway through, but I think this should be good now.

Manifest.toml
davidanthoff: I'm a big believer in not including `Manifest.toml`s in packages :) In my mind they record things that shouldn't be recorded in a package.

Not sure I understand the upper bound scenario.

staticfloat: > I'm a big believer in not including Manifest.tomls in packages :)

It's your package, so by all means, manage it how you like!  I'll remove it.

> Not sure I understand the upper bound scenario.

The `Manifest.toml` records a "known good point" within the entire space of packages that the resolver can choose.  If you use `Pkg.add("Cairo")`, you do not necessarily get a working setup for `Cairo.jl`, since the resolver is free to choose among a potentially very wide number of packages.  If the dependency upper bounds are chosen perfectly, `Pkg.add()` should always work.  However, it's pretty common that somebody updates a package somewhere in your deps tree that breaks something higher up in the deps tree, eventually causing `Cairo.jl` to break.  This doesn't even have to be `Cairo.jl`'s fault; it could be a transitive dependency that causes the problem.

Bundling a `Manifest.toml` provides two tools to help out:
* If you are in the "just work, darnit" frame of mind, you can instantiate the `Manifest.toml` to get the versions that worked in the past (and thereby should always work, especially in light of JLL packages which extend instantiation's effect to binary packages).
* If you are trying to debug what versions of a dependency _should_ work, the `Manifest` at least gives you a known good point; you don't need to go and look at the Registry to see what versions would have been installed on your machine six months ago when you last modified this package; all that information is stored here.

Manifest.toml files bundled with packages aren't used for resolution or loading in any way during normal usage; you have to explicitly start Julia with `--project=` pointing at `Cairo.jl` for them to be used, and they'll only install their versions if you explicitly `instantiate`.

davidanthoff: > It's your package, so by all means, manage it how you like! I'll remove it.

Actually, it is not :) Not sure who is in charge, probably @lobingera?

Thanks for the explanation of the benefits, interesting. I think one main reason why I don't like to have `Manifest.toml`s in my packages is that VS Code will auto-activate an env if there is a `Manifest.toml` in a package folder, and that is something I don't want, so it throws a wrench into my normal workflow. Also, it just seems that almost all packages don't have it in the root, and I'm a fan of just following common practices with this kind of choice :)

staticfloat: Ah, I had no idea VS Code did that.  Interesting.  I can see the reason it does it, but yes that would be frustrating.  This is where the concept of an "application" versus a "package" start to conflict a little bit.  :P

davidanthoff: From the VS Code extension point of view, it would be _great_ if a package project was called `Package.toml`, and an app project was called `Project.toml`, even if they have exactly the same syntax/content. Or some other way to distinguish a package from an app. It would just be really convenient for tools to be able to look at a folder and just say "ah, this is a package", or "ah, this is an app"...

tknopp:
To move forward I merged this now. Will prepare the next release if there are no objections from @lobingera. Regarding version number. I don't care but I do also not see the pressing advantage of 1.0.0. Currently any bump of the 0.x is breaking. Afterwards any change of the first version number is breaking. I am fine with both.

Issue #294, opened 2019-09-06
New release version number?


timholy:
Gtk is currently incompatible with the master branch here, but there is little reason to avoid releasing this as long as we impose adequate version bounds. So let's discuss: what version number should we assign? I would personally favor 1.0, but happy to hear others' thoughts (esp. @lobingera as the maintainer!)

lobingera:
I'm too conservative on things like this, i'd apply a 0.7 (or even smaller)

I think i mentioned this some otherplace: For me a 1.0 should be step into a modular Cairo  + Pango + FreeType (optional dependencies or glue packages), a second look at Graphics.jl, MORE testing, much more coverage of libcairo C-APIs (like https://www.cairographics.org/manual/cairo-text.html and e.g. https://www.cairographics.org/manual/cairo-Raster-Sources.html). And as i recently looked into Clang.jl and types for ccalls some work on memory management (#270 #225).

timholy:
OK, I'll assume it will be 0.7 and bound Gtk accordingly. (It should not be 0.6.1 just in case it ends up causing problems on certain platforms, we want some room to release a few more 0.6.x versions if necessary.)

davidanthoff:
I'd be in favor of 1.0.0, mainly because it moves us into "normal" semver territory where we can use different version numbers to clearly indicate bug fixes, new features and breaking releases. For the longest time I was in the camp of "lets wait until we have something stable until a 1.0.0", but I've very much changed my view with the new package manager: it is just a heck of a lot easier for downstream packages to put proper bounds onto things if a package is post 1.0. So I just don't attach any special meaning to 1.0.0 anymore, but mainly treat it as unlocking features in the package manager resolver.

tknopp:
This is definitely not 0.6.1. Regarding 0.7 or 1.0, I do not care too much but somewhat agree with @davidanthoff.

PR #297, opened 2019-09-14
Version 0.7.0

timholy:
I went with 0.7 from #294 given @lobingera's preference, but I think @davidanthoff's point is good and would be happy to change this to 1.0.


cormullion:
Either would be great! :)

PR #301, opened 2019-10-03
Bump dependencies to eradicate symlinks

staticfloat:
This should help installations problems on Windows


davidanthoff:
Can we also tag a release?

PR #303, opened 2019-10-11
Update binaries

davidanthoff:
This updates the binaries we are using to versions that work with a BP based Rsvg.jl as it is in https://github.com/lobingera/Rsvg.jl/pull/35.

I increment the version of the package here because I think we should treat changes like this as breaking.


davidanthoff:
Alright, so the big three all seem to work, but FreeBSD and the drone build don't. I'm a bit at a loss what is going on there. @staticfloat and @giordano, any idea?

davidanthoff:
Hurray, it all seems to work!

@lobingera could you merge and then tag a release, v0.8.0? Once that is done, we should be able to finish the Rsvg.jl conversion to BinaryProvider as well :)

deps/build_Pango.v1.42.4.jl
```diff
@@ -46,4 +47,4 @@ if unsatisfied || !isinstalled(dl_info...; prefix=prefix)
 end
 
 # Write out a deps.jl file that will contain mappings for our products
-write_deps_file(joinpath(@__DIR__, "deps.jl"), products, verbose=verbose)
+write_deps_file(joinpath(@__DIR__, "deps.jl"), products, verbose=verbose)
```
giordano: Missing newline here :slightly_smiling_face:

deps/build_HarfBuzz.v2.6.1.jl
```diff
@@ -44,4 +47,4 @@ if unsatisfied || !isinstalled(dl_info...; prefix=prefix)
 end
 
 # Write out a deps.jl file that will contain mappings for our products
-write_deps_file(joinpath(@__DIR__, "deps.jl"), products, verbose=verbose)
+write_deps_file(joinpath(@__DIR__, "deps.jl"), products, verbose=verbose)
```
giordano: Missing new line

davidanthoff:
Thanks! @lobingera can you also tag a new release?

lobingera:
@JuliaRegistrator register

PR #305, opened 2019-10-13
CI update

davidanthoff:
This updates our general CI config. Changes are:
- Use travis instead of appveyor for testing on Windows. Seems faster and one less moving part.
- Run tests on Julia 1.1. I think if we declare compat with that version, we should test it.
- Use the `coveralls` flag in travis instead of the manual code version.

This should probably go hand-in-hand with deleting the appveyor project.


giordano:
Wow, segmentation faults :slightly_smiling_face:

davidanthoff:
And we don't seem to have a julia 1.1 image for the drone build? Who is maintaining these, and is there a chance that we can add that?

giordano:
They removed Julia 1.1 while upgrading to 1.2: https://github.com/docker-library/julia/commit/5f637f20522f82f953c69318002820da70d2e3f9  I guess because 1.1 is unmaintained

PR #309, opened 2019-12-08
Return `status_t` from `write_to_png`

timholy:


timholy:
The aim here would be to be able to debug the test failure in https://travis-ci.org/JuliaGizmos/GtkReactive.jl/jobs/622293709?utm_medium=notification&utm_source=github_status, which only happens on Travis.

timholy:
Bump

lobingera:
Meaning? I thought you're juliagraphics member, so you can merge as you like

timholy:
I can, I just wanted to make sure you didn't disapprove.

PR #317, opened 2020-04-17
Support Colors 0.12

timholy:
I've tested this locally and it passes tests. In the Travis logs it's worth checking whether it actually uses this version, though.

We might want to install https://github.com/bcbi/CompatHelper.jl, though https://github.com/bcbi/CompatHelper.jl/issues/160 is pretty annoying. But since Cairo's dependencies are pretty minimal, I doubt it will be that much of a problem here (it comes up much more often in packages with lots of dependencies).


timholy:
Allowed failure on nightly is fixed already by ColorTypes 0.10.2.

PR #318, opened 2020-04-17
Create CompatHelper.yml

timholy:
https://github.com/bcbi/CompatHelper.jl

### DIFF
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -1,0 +1,16 @@
+name: CompatHelper
+
+on:
+  schedule:
+    - cron: '00 00 * * *'
+
+jobs:
+  CompatHelper:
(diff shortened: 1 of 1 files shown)


giordano:
Can you please retrigger the drone jobs that failed on 32-bit ARM?  It should be fixed now.

timholy:
I don't think I can, but perhaps @lobingera?

timholy:
Actually, let's try a close/open cycle.

lobingera:
@timholy @giordano 

I have lost track, is this needed?

giordano:
It's not _needed_, but it helps keeping the compat bounds in `Project.toml` up-to-date: a GitHub actions scheduled every day will open a pull request, if necessary, to update them.  In this way you don't need to manually go through each dependency and check whether there are new releases.

lobingera:
Merged.

PR #325, opened 2020-07-04
Fix stream handling

### LINKED ISSUE #323: segfaults on v1.5.0-beta1
The following line seems to be causing segfaults in Julia version v1.5.0-beta1 (see https://github.com/JuliaPlots/Makie.jl/issues/627):
https://github.com/JuliaGraphics/Cairo.jl/blob/25f5658c9b70075bd3f1a781febc5a38c5cbd002/src/Cairo.jl#L158

Besides, it seems strange to predicate on a literal `false`?

yuyichao:
Passing immutable object as `Any` to C code does not guarantee pointer identity
with the original object/object stored in parent object.

Fix #323

### DIFF
--- a/src/Cairo.jl
+++ b/src/Cairo.jl
@@ -222,6 +222,11 @@ mutable struct CairoSurfaceIOStream{T<:Union{UInt32,RGB24,ARGB32}} <: CairoSurfa
         self
     end
 end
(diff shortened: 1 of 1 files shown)


lobingera:
Thank you. 

I still might have some questions understanding the solution.

Issue #330, opened 2020-08-02
Access current font size


Wikunia:
It's currently not possible to get the current font size, right?
I assume it's possible to do this with: https://www.cairographics.org/manual/cairo-text.html#cairo-get-font-matrix
Maybe someone can help me with creating a PR for that. 
Unfortunately I haven't worked with Cairo or `ccall` before :smile: so I'm a bit lost

lobingera:
It should be very close to 
https://github.com/JuliaGraphics/Cairo.jl/blob/master/src/Cairo.jl#L1096-L1119
which is the operations on the context's matrix.

PR #331, opened 2020-08-02
set/get_font_matrix

Wikunia:
Attempt to implement #330 
I think font size is something Luxor/@cormullion might be interested in.


Wikunia:
Is it possible to get this tagged such that https://github.com/JuliaGraphics/Luxor.jl/pull/104 can be continued? Thanks in advance

lobingera:
I can do it, but not before sunday

lobingera:
https://github.com/JuliaRegistries/General/pull/19442

PR #334, opened 2020-10-29
Provide macro @libcall to simplify ccall's

emmt:
A macro is provided to automatically use the correct library based on the prefix of the function. This makes the code a bit more readable and less prone to errors.  The macro `Cairo.@libcall` can be used in other packages to simplify calling a Cairo/Pango function.


lobingera:
"and less prone to errors" - which errors?

emmt:
Well when the code is more clear and easier to read, I think that it less likely to make it wrong. I did not meant that that there were errors ;-)

asinghvi17:
This now exists in Base's `@ccall` macro.

ViralBShah:
Agree this should be done with `@ccall`

PR #337, opened 2020-12-28
Switch CI to GitHub Actions

giordano:
Also update TagBot and CompatHelper setups.

In order to build the documentation you'll need to set up the [Documenter SSH key for GitHub Actions](https://juliadocs.github.io/Documenter.jl/stable/man/hosting/#travis-ssh).  You may also want to disable running AppVeyor and/or deleting its webhook


lobingera:
Race condition ... i'm testing this currently in a different place ...

lobingera:
but you are right, appveyor should go and we don't have documentation that's build by Documenter.

giordano:
> we don't have documentation that's build by Documenter.

Ugh, I hadn't checked that.  Removed

PR #338, opened 2020-12-31
Make height and width inferrable for (abstract) CairoSurface

timholy:
Gtk.jl and other libraries have containers with fields that are
typed abstractly as `::CairoSurface`. This allows `width` and
`height` to be inferrable even for the abstract type.
It enforces an "interface" expected of all CairoSurface subtypes.

### DIFF
--- a/src/Cairo.jl
+++ b/src/Cairo.jl
@@ -169,6 +169,18 @@ get_readstream_callback(::Type{T}) where T = @cfunction read_from_stream_callbac
 
 abstract type CairoSurface{T<:Union{UInt32,RGB24,ARGB32}} <: GraphicsDevice end
 
+# All CairoSurfaces have to implement at least the fields and types of CairoSurfaceBase
+function Base.getproperty(surface::CairoSurface, fieldname::Symbol)
+    if fieldname === :ptr
+        return getfield(surface, :ptr)::Ptr{Nothing}
+    elseif fieldname === :width
+        return getfield(surface, :width)::Float64
+    elseif fieldname === :height
+        return getfield(surface, :height)::Float64
+    end
+    return getfield(surface, fieldname)
+end
+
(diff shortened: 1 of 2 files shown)


timholy:
@lobingera, if you're fine with saying goodby to appveyor, don't forget to delete the webhooks (Settings -> web hooks).

lobingera:
LGTM (although i wonder what the usecase is ...), web-hooks to appveyor deleted.

timholy:
> although i wonder what the usecase is 

Marginally better performance with Gtk (since Gtk doesn't store *which* kind of `CairoSurface` it is), but the real benefit is better precompilability and thus lower latency on first use. See https://github.com/JuliaLang/www.julialang.org/pull/1111 and a preview at https://julialang.netlify.app/previews/pr1111/blog/2021/01/precompile_tutorial/.

This is a very small change and thus it has only small impact, but it's something I noticed while engaged in tracking down sources of inference failures in my soon-to-be-released [GtkObservables](https://github.com/JuliaGizmos/GtkObservables.jl) package.

timholy:
Oops, though I see I messed up this test...fixed in #339.

lobingera:
I see. But; as far as i understand, the width and height stored in the surface type are more informative and the dimensions of create surfaces - not necessariiy how the data is arranged in memory (see stride information in https://www.cairographics.org/manual/cairo-Image-Surfaces.html). And your comment about "Gtk doesn't store which kind of CairoSurface it is" - i don't think Gtk offers more than the CairoContext in case of widget (re-)drawing and afair in cases of redrawing you get clipping, but not more information. So deriving width+height from Context -> Surface for rendering on screen (and your definitions for precomp for user_to_device is another indication...) might work in some cases, but not in the general case with e.g. overlapping windows.

Issue #355, opened 2022-08-20
Add wrapper for cairo_recording_surface_ink_extents()


hustf:
This is useful to compute the required size of an image surface (or equivalent) into which to replay the full sequence of drawing operations.

Api reference, since version 1.10:
https://www.cairographics.org/manual/cairo-Recording-Surfaces.html#cairo-recording-surface-ink-extents

Use case with Luxor (this helps to fix):
https://github.com/JuliaGraphics/Luxor.jl/issues/150.

PR #356, opened 2022-08-20
Add wrapper for cairo_recording_surface_ink_extents()

hustf:
Ref.  #355 

modified:   samples/Samples.md Include sample_record_inkextents
new file:   samples/sample_record_inkextents.jl
new file:   samples/sample_record_inkextents.png
modified:   src/Cairo.jl    Add ink_extents


hustf:
As per @cormullion comment elsewhere: The only test so far follows the existing setup: A 'sample' file is included in 'runtests'. The test checks that an image file was produced. We could extend this to check pixel counts, but I am not sure if that would be a good test of weak points in this case. Difficult tests would include 
* large variation of scales
* extending the canvas from positive to negative
* drawing shapes with non-trivial bounding boxes

But in these cases, we would be testing Cairo itself, not this wrapper.

PR #357, opened 2022-10-03
Allow to use FreeTypeAbstraction fonts and render glyph by index

Kolaru:
These are small features that I am porting from CairoMakie so that they will be available to Luxor as well, to allow support for MathTeXEngine v0.5.

I do not know if I should add tests for those or what those tests should look like.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -16,6 +16,7 @@ Cairo_jll = "1.16.0"
 Colors = "0.9, 0.10, 0.11, 0.12"
 Glib_jll = "2.59.0"
 Graphics = "0.4, 1"
+FreeTypeAbstraction = "0.10"
 Pango_jll = "1.42.4"
 julia = "1.3"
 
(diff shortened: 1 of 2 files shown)


cormullion:
@lobingera How does this look to you?

lobingera:
Hmm, I lost track how we do things in julia today. Code looks reasonable and i think i have a local branch that does more or less the same, but i never saw libfreetext working correctly on all platforms.

Testing - just calling all functions - would be nice. Having a standard font available that loads in all platforms is a plus. Doing an example is the highest i can ask for.

davibarreira:
This PR is greatly appreciated!

ViralBShah:
Please let me know if this should be merged? Also happy to give access to a couple of more folks to maintain the package.

cormullion:
> Please let me know if this should be merged? Also happy to give access to a couple of more folks to maintain the package.

It would be great if Cairo.jl has occasional maintainers - it's pretty stable and unchanging generally.

I think Cairo.jl is still fairly central to the Julia ecosystem - probably needs releases and updates occasionally.

This PR probably needs some tests...

ViralBShah:
Good point. Maybe we should see if any of the folks who have made a couple of PRs to Cairo.jl are interested in stepping up for light maintenance.

PR #361, opened 2024-02-26
Bump Cairo_jll compat to the latest version

asinghvi17:


asinghvi17:
Cairo_jll's minimum Julia compat for v1.18 is Julia 1.6, which is why the Julia 1.3 CI is failing to install it...

giordano:
Well, you'd get the same without this PR.

Issue #363, opened 2024-04-01
Cairo/Windows ReadOnlyMemoryError()


cormullion:
When running my tests on CI (both `Julia 1.9 - windows-latest - x64` and `Julia 1 - windows-latest - x64`) I'm seeing a `ReadOnlyMemoryError()`:

```plain
 Got exception outside of a @test
  LoadError: ReadOnlyMemoryError()
  Stacktrace:
    [1] get_layout_size
      @ C:\Users\runneradmin\.julia\packages\Cairo\smWIA\src\Cairo.jl:1184 [inlined]
    [2] text(ctx::Cairo.CairoContext, x::Float64, y::Float64, str::String; halign::String, valign::String, angle::Int64, markup::Bool)
      @ Cairo C:\Users\runneradmin\.julia\packages\Cairo\smWIA\src\Cairo.jl:1265
    [3] text
      @ C:\Users\runneradmin\.julia\packages\Cairo\smWIA\src\Cairo.jl:1255 [inlined]
```

[from log here](https://github.com/JuliaGraphics/Luxor.jl/actions/runs/8506130997/job/23295828666#step:6:455)

I don't know what this error means or how significant it is... Tests run OK on MacOS and Linux. Cairo v1.0.5.

mwlidar:
Had the same problem. What helped was to revert (and pin) Glib_jll to version 2.76.5.
The downside is that this is in conflict with the latest versions of Gtk4 and CairoMakie...

There are also other packages which run perfectly fine on non-Windows platforms but suffer from these ReadOnlyMemoryErrors on Windows, like for example GRIB.jl. I really hope that a Windows-expert will popup soon to solve this!

jwahlstrand:
For me locally this is fixed with Pango_jll 1.54.0.

mwlidar:
For me the crashes (1.10.4 / Windows11) are also gone with the latest versions of Cairo/Pango etc.
But now I have text cut off by a few pixels at the top 😒
Will open a separate issue on that...

Issue #349, opened 2021-12-03
Pango problems with smaller fonts


mwlidar:
Since a few day smaller fonts (below 16pt) are not rendered correctly any more. Most easyly to see with the sample_pango_text.jl example:
![sample_pango_text](https://user-images.githubusercontent.com/87604697/144637779-ffaf0719-d409-4747-a06e-be435bc80f32.png)

This is with Julia 1.6.4 and Cairo 1.0.5. Must have to do with the lates FreeType package updates. But I wasn't able to narrow this down further...

hustf:
It seems this affects all font scaling in Pango. If that is the case, a new version ought to be coming soon.

```text(cr,16.0,104.0,"Text<b>Bold</b><i>Italic</i><big>super-2</big>",markup=true)```

Confirmed on Windows, Julia 0.7.0.

```
julia> libpango_version = unsafe_string(
             ccall((:pango_version_string,Cairo.libpango),Cstring,()) )
"1.47.0"
```

hustf:
Another example, indirectly to Cairo using Luxor. Font scaling is broken (with my setup).
It would be nice to know if this is an issue with the windows version, only? 

´´´
using Luxor
Drawing(640, 480, "temp.svg")
settext("Before, ϕ", Point(100.0, 150.0))
setfont("Georgia Bold", 65.0)
settext("Before, ϕ", Point(100.0, 200.0))
finish()
preview()
´´´

In case issues get posted at Luxor.jl, for info: @cormullion

hustf:
'One size per font family per session' is close to describing the issue here.

It's daunting that the error remains until the next Julia session. Also, is this an issue on Windows only?

![sample_pango_text2a](https://user-images.githubusercontent.com/13949620/145845073-b4b218b0-978b-4be2-ae44-ccce892888a1.png)

![sample_pango_text2b](https://user-images.githubusercontent.com/13949620/145845141-fad20dc6-509a-4434-b8de-8e5d7e708cf0.png)

´´´
using Cairo
c = CairoRGBSurface(256,256);
cr = CairoContext(c);
fill(cr);
set_source_rgb(cr,0.8,0.8,0.8);    # light gray
set_font_face(cr, "Sans 16.0")

text(cr,40.0,60.0,"Sans 16.0")
text(cr,40.0,80.0,"Sans 16<small>small</small>",markup=true)
set_font_face(cr, "Sans 10.0")
text(cr,40.0,100.0,"Sans 10")
[...]

PR #365, opened 2024-08-12
Update version to 1.1 in Project.toml

ViralBShah:
Bumping up dependencies for very old versions. Perhaps this can be done more aggressively.

Does someone have views on Cairo, Pango, and Glib? In general those builds made a lot of improvements in the last few versions and it may be nice to set a new minimum. Example: https://github.com/JuliaGraphics/Cairo.jl/issues/363#issuecomment-2220273451

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Cairo"
 uuid = "159f3aea-2a34-519c-b102-8c37f9878175"
-version = "1.0.5"
+version = "1.1"
 
 [deps]
 Cairo_jll = "83423d85-b0ee-5818-9007-b63ccbeb887a"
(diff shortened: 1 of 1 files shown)


ViralBShah:
@jwahlstrand Hope you don't mind the ping. Do you have any thoughts on bumping the minimum versions of Cairo etc. here (whether we should or should not)?

cormullion:
@ViralBShah Makes sense to me.

ViralBShah:
Merging for now, but will wait before registering, in case others are able to chime in.

jwahlstrand:
I think continuing to allow Pango_jll 1.42 is a good idea because of https://github.com/JuliaGraphics/Cairo.jl/issues/349, which last I checked is still a problem on Windows. The workaround is to downgrade Pango_jll to that version. Updating Harfbuzz_jll (https://github.com/JuliaPackaging/Yggdrasil/pull/8392) fixes it but there is a problem with the MacOS build.

I think it's fine to leave Cairo_jll and Glib_jll minimum versions as they are for now. None of the new API in libcairo 1.18 is wrapped yet.

PR #370, opened 2024-12-24
wrap `set_metadata` and `set_custom_metadata`

### LINKED ISSUE #360: Metadata for pdfs - `cairo_pdf_surface_set_metadata()`
Would it be possible to add a wrapper for `cairo_pdf_surface_set_metadata()` (https://www.cairographics.org/manual/cairo-PDF-Surfaces.html#cairo-pdf-surface-set-metadata)?

I think this would be useful to enable https://github.com/MakieOrg/Makie.jl/issues/3167.

I would be willing to have a go at this, but have no experience using `ccall`s or developing `Cairo.jl` - would it be quicker for someone who knows what they're doing to just add this than it would be to review a PR?

jwahlstrand:
Wrap these two methods for setting metadata in PDF surfaces. Also adds status enums that were added in cairo 1.16 and 1.18. Includes tests. Also bumps required Cairo version to 1.18.

Fixes https://github.com/JuliaGraphics/Cairo.jl/issues/360

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -11,7 +11,7 @@ Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
 Pango_jll = "36c8627f-9965-5494-a995-c6b170f724f3"
 
 [compat]
-Cairo_jll = "1.16"
+Cairo_jll = "1.18"
 Colors = "0.12, 0.13"
 Glib_jll = "2.59.0"
 Graphics = "1"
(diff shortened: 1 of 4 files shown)


ViralBShah:
@jwahlstrand Seems good to merge (even though old PR). Shall we go ahead? If you don't have commit access here, happy to add you!

jwahlstrand:
> @jwahlstrand Seems good to merge (even though old PR). Shall we go ahead? If you don't have commit access here, happy to add you!

Yes, I think this is worth merging, thanks! If I had commit access here I'd be willing to do light maintenance going forward.

